Issue #22, opened 2020-10-28
server_default documentation examples and read-only field


HarrySky:
Hey, found this library through your comment in `orm` repo, want to try it, as it seems it is not abandoned :smile:

I want to set field that is updated by PostgreSQL triggers, not directly, so I thought that `server_default` would be the option I need to pass to just ignore the field when creating/updating object. But I found no examples in documentation that I can just grab and try.

So, my field is defined in database as:
```sql
dt_changed timestamp with time zone NOT NULL DEFAULT NOW()
```

How can I pass `NOW()` function to `server_default`?

Related feature proposal: Can we introduce something like "read-only" field (so that it does not try to create/update it)?

Can submit PR for documentation and/or "read-only" field if you like the idea

collerek:
Hi,

I remember messing around with `server_default` when I was setting `pydantic.ModelField` default values so I conducted some tests and fixed the `server_default` parameter and in the so called mean time updated the documentation in this regard.

Was fixed in #26, please update and check.

So answering your question directly you need to pass `sqlalchemy.func.now()` to server_default to achieve this functionality.

As for the `read_only` - I assume you want something like `server_default` field that cannot be updated.

Some requirements/ specs to consider:
*  If a field is read_only it has to have server_default set (maybe also allow default?). Otherwise it will be useless if cannot be used during creation/update
* Field need to be excluded from query in model save/update but also in QuerySet level create, update etc. ManyToMany models also expose QuerySet API but it's delegated to actual QuerySet later so that should be fine.
* Should we just exclude the field or raise Exception if someone wants to update/set read_only field?


Sure might be useful as some kind of 'security' feature to not allow overwriting the field, if you wish to create a PR it will be welcomed.

HarrySky:
Thank you, will check tomorrow and re-open if something does not work :smile:

About `read_only` - yes, if I have triggers on PostgreSQL side - there is no need to write field value on update as it will be overwritten, example:
I have a `dt_changed` on model, it is set to "5 seconds ago" when I create or load object, but will it try to write this "5 seconds ago" value on subsequent updates? If it will not try, then no writes would be wasted and everything is ok :+1: 

Though it was more for cases when you already have a database and do not create schema through Python.

collerek:
If you provide a value both `default` and `server_default` are ignored so if you write model to db first you will get model with this field populated so for any subsequent call the value provided will be used. If you save again this value will be passed to updates so any function will not run, but the column will be included in update query. 

[...]

PR #26, opened 2020-10-28
Fix server_default Field parameter 

collerek:
*  Fix parameter server_default being set on create not on server side
*  Exclude None fields with server_default to not overwrite the default
*  Update docs as requested in #22


HarrySky:
Thanks a lot! :rocket:

PR #29, opened 2020-11-03
Fix inconsistencies in documentation

HarrySky:
- Use max_length instead of length for String field
- Use same annotations in README.md and index.md as in docs_src


collerek:
Thanks for the updates! 🙂

PR #40, opened 2020-11-11
Fixed line ordering in install.md

smorokin:
Two lines were mixed up


collerek:
Thanks for catching that up! :)

smorokin:
Thanks for merging that so fast :-).

Issue #50, opened 2020-11-22
Allow saving /adding /removing related models from reverse side of ForeignKey relation


collerek:
Something more in line with ManyToMany field behavior without saving through models

collerek:
Closed in #61

PR #52, opened 2020-11-24
support updating models on the many side of the one-to-many relations.

EspenAlbert:
Hi.
I ran into some issues updating models on the "many" side of one-to-many relations.
E.g., Album with many tracks.
As seen in the new test:
`await Album.objects.bulk_update([album], columns=["name"])`

This PR attempts to fix those issues :)

However, it might be that I am using the api wrong.
I don't know why the tracks are not updated as seen in the last test:
`assert len(updated_album.tracks) == 0`

I think this is related to https://github.com/collerek/ormar/issues/50

### DIFF
--- a/ormar/models/modelproxy.py
+++ b/ormar/models/modelproxy.py
@@ -70,14 +70,15 @@ class ModelTableProxy:
                             f"model without pk set!"
                         )
                     model_dict[field] = pk_value
-                elif isinstance(field_value, list):
-                    targets = [target.get(target_pkname) for target in field_value]
-                    if targets:
-                        model_dict[field] = targets
+                elif field_value:  # nested dict
+                    if isinstance(field_value, list):
+                        model_dict[field] = [
+                            target.get(target_pkname) for target in field_value
+                        ]
                     else:
-                        model_dict.pop(field)
-                else:  # nested dict
-                    model_dict[field] = field_value.get(target_pkname)
+                        model_dict[field] = field_value.get(target_pkname)
+                else:
+                    model_dict.pop(field)
         return model_dict
 
     @classmethod
--- a/ormar/models/modelproxy.py
+++ b/ormar/models/modelproxy.py
@@ -78,7 +78,7 @@ class ModelTableProxy:
                     else:
                         model_dict[field] = field_value.get(target_pkname)
                 else:
-                    model_dict.pop(field)
+                    model_dict.pop(field, None)
         return model_dict
 
     @classmethod
--- a/ormar/models/modelproxy.py
+++ b/ormar/models/modelproxy.py
@@ -70,6 +70,12 @@ class ModelTableProxy:
                             f"model without pk set!"
                         )
                     model_dict[field] = pk_value
+                elif isinstance(field_value, list):
+                    targets = [target.get(target_pkname) for target in field_value]
+                    if targets:
+                        model_dict[field] = targets
+                    else:
+                        model_dict.pop(field)
                 elif field_value:  # nested dict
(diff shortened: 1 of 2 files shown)


collerek:
Hi, it's hard to say what you try to achieve from this snippet :) Please provide more complete example.

But by definition `bulk_update` and `bulk_create` affects only the directly related models (so the ones in the list), it does not follow any relation. So it will update `ForeignKey` relation as the FK is actually kept on the model side, but will not touch `ManyToMany` and second side of `ForeignKey` (virtual FK).

So if you update `Album`, the `tracks` will not be updated as it's a second side of FK relation (I assume it's the same as in examples and tests). So the column is actually kept on `Track` side.

That's why `save_related` was introduced. It's a known "problem" that was considered by `marshmallow`, `pydantic` and similar libraries. Cause it's perfectly valid (and actually `ormar` is doing the same) to keep a relation like `ModelA -> ModelB -> ModelA` -> so a circular relation. And the decision when to stop loading/updating the relation is arbitral. (cause if you go `ModelA -> ModelB -> ModelA -> should you load ModelB again`?).

The `save_related` updates each related model only once. So even if you have `ModelA -> ModelB -> ModelA -> ModelD` - since the `ModelA` was the first model it won't ever go to `ModelD`, as the second `ModelA` occurrence is already duplicated. So If you try to save nested models like this you need to do it yourself beforehand.

To update deep nested models you need to provide `save_related(follow=True)` as by default `save_related` saves only directly related models. So in example `ModelA -> ModelB -> ModelC` it will save only models from `ModelB` and do not go deeper.

I remember there was a check for a list but later it became obsolete and I removed it. 
But even if it's actually needed you should go with something like this:

```python
elif field_value:  # both empty list and dict will evaluate to False
    if isinstance(field_value, list):
        model_dict[field] = [target.get(target_pkname) for target in field_value]
    else:
        model_dict[field] = field_value.get(target_pkname)
else:  # if it's empty list or dict you need to set it to None 
    # as otherwise pydantic will try to instantiate nested models and fail
    # even if the field is optional
    model_dict.pop(field, None)
```

As it's actually a one check less (cause regardless of the fact if the dict or list is empty it;s the same if, so lower cyclomatic complexity).

EspenAlbert:
Thanks for the great explanations. I really appreciate it :)

I tried to improve the test by adding `play_count` to tracks and `is_best_seller` to albums.
But still, it might not be that good to explain the point. (But I didn't want to include the business case I am dealing with as it is a bit overkill...)

But the main problem is updating any model that is used on the "other" side of a `ForeignKey`

ormar/models/modelproxy.py
```diff
@@ -71,9 +71,14 @@ def substitute_models_with_pks(cls, model_dict: Dict) -> Dict:  # noqa  CCR001
...
-                    model_dict[field] = field_value.get(target_pkname)
+                    if isinstance(field_value, list):
+                        model_dict[field] = [
+                            target.get(target_pkname) for target in field_value
+                        ]
+                    else:
+                        model_dict[field] = field_value.get(target_pkname)
                 else:
-                    model_dict.pop(field, None)
+                    model_dict.pop(field)
```
collerek, line 81: Please leave the `pop(field, None)` -> the whole idea here is to remove the field from dict if it's empty in data. So if the field is missing/not there for whatever reason it's ok - we don't have to remove it, but if the field is missing popping it will cause exception.

collerek (changes requested):

collerek (approved):

collerek:
Thanks for the PR and welcome among the contributors/creators :)

Issue #112, opened 2021-02-22
Enhance the TypeHints to return actual Models and not ormar.Model


collerek:
As described in #111 `QuerySet` methods should actually return inferred Model subclasses and not the main `ormar.Model`.

Since I am not an expert in complex python typing any help appreciated.

aliakhtar:
Yes, great library, but can we please have a release that fixes this issue? Hopefully it also works for `objects.get()`

collerek:
There is already PR #132  that should fix all `objects.method()` calls, which is not all that is to fix, but the most common calls should be handled.

aliakhtar:
Thanks - is that available on pip?

collerek:
If you mean the PR, not yet, need to finish the docs and will merge and publish this probably today. 

As of now you can install from source (gh branch) I guess.

aliakhtar:
Awesome - Love it!

I think Model.objects.get() should return an optional instead of raising a NoMatch - do you agree @collerek ? If so I can do a PR for that.

collerek:
@aliakhtar As of now I prefer the exception, since `None` is a perfectly valid return type, especially for aggregated functions like `max` or `min`. Without exception you don't know if it's an actual value from the database or there is no match for your query criteria. 

Following pep8: `Explicit is better than implicit`, and for me returning `None` where there is no record in db with regard to ORM is kind of an implicit error silencing, and it makes testing and debugging more difficult (did I make a typo and there is no record or something was saved in a wrong way and the data is corrupted etc.)

If you don't like it you can very easily write a `decorator/wrapper/helper` that will silence this error and return `None`, but other way around (getting to know if it's db value or error) would be pretty hard without subclassing and overwriting implementation (which of course you can also do).

collerek:
@aliakhtar that's not so simple as they rely on the same exception :) 

Moreover this would be a breaking change that would have to wait for next bigger release, to add even more I like the exception approach more and there is quite some amount of users that do too.

But since it's not the first time someone asks me for this functionality I see a place for this but more like a setting then a permament change. I can add a setting on Model's Meta class (that can be easily shared by all models through `BaseMeta` approach if you use one database only (check best practices in docs)), something like `exception_on_no_match=True` that you can change to `False`. 

Can you please open an enhancement issue for this?

aliakhtar:
What about a new method called `get_if_exists` with the type signature to
return optional instead? Happy to send a PR for it.

[...]

PR #124, opened 2021-03-05
#112 - Resolve MyPy Errors

dudil:
Following discussion there is only 1 error left which is related to Return "Model" from Model methods.
This is possible solution for that error:
https://stackoverflow.com/questions/15853469/putting-current-class-as-return-type-annotation


collerek:
It fails on runtime as you define `TM` import in `TYPE_CHECKING` parts and `TM` and `TypeTM` does not exists during runtime.
It can be fixed by wrapping types in quotes but that won't work for `Generic[TM]` as type passed to generic cannot be a string.
So either you need to redefine it in files (but that might cause different problems of type mismatching), or use ormar.TM to avoid circular imports.

Apart from `mypy ormar tests` you should run `pytest tests`. Tests are also run on all 3 backends when you issue a PR.

dudil:
Got you - Will look into it now :)

dudil:
Hi @collerek - 

It seems the change you did on model.py broke the behavior of version 0.9.6
The usage of generic here is not working as intended and when trying with my live code using the latest package it is mixing Models not as intended.

Maybe best to roll back this change as well?
It is better to handle it as a dedicated branch I believe

collerek:
Can you provide an example, cause i actually restored some of the type hints that were already there in 0.9.5 and earlier but not all of them

dudil:
I'm sorry - I think I got confused with all the changes.
The example I had in mind is not a good one and I think it is more confusing than helping now.

If you still want me to assist with this issue and make the same change I did without breaking the tests - please let me know and I'll happy to do so!

dudil:
Great - will be happy to assist.
Feel free to let me know what is your plan and how I can be at your assistance with that.

collerek:
I created types_enhance branch and play with it to make it work.

There is a test file in tests category: test_types.py that has bunch of code to reveal types during types check.

```
query = Book.objects
publisher = await Publisher(name="Test publisher").save()
author = await Author.objects.create(name="Test Author")
await author.publishers.add(publisher)
author2 = await Author.objects.select_related("publishers").get()
publishers = author2.publishers
publisher2 = await Publisher.objects.select_related("authors").get()
authors = publisher2.authors
assert authors[0] == author
for author in authors:
    if TYPE_CHECKING:  # pragma: no cover
        reveal_type(author)  # iter of relation proxy
book = await Book.objects.create(title="Test", author=author)
book2 = await Book.objects.select_related("author").get()
books = await Book.objects.select_related("author").all()
author_books = await author.books.all()
assert book.author.name == "Test Author"
assert book2.author.name == "Test Author"
if TYPE_CHECKING:  # pragma: no cover
    reveal_type(publisher)  # model method
    reveal_type(publishers)  # many to many
    reveal_type(publishers[0])  # item in m2m list
    # getting relation without __getattribute__
    reveal_type(
        publisher2._extract_related_model_instead_of_field("authors")
    )  # TODO: wrong
    reveal_type(authors)  # reverse many to many  # TODO: wrong
    reveal_type(book2)  # queryset get
    reveal_type(books)  # queryset all
    reveal_type(book)  # queryset - create
    reveal_type(query)  # queryset itself
    reveal_type(book.author)  # fk
    reveal_type(
        author.books._queryset
    )  # queryset in querysetproxy  # TODO: wrong
    reveal_type(author.books)  # reverse fk  # TODO: wrong
    reveal_type(author)  # another test for queryset get different model
    reveal_type(book.author.name)  # field on related model
    reveal_type(author_books)  # querysetproxy for fk  # TODO: wrong
    reveal_type(author_books[0])  # item i qs proxy for fk  # TODO: wrong
assert_type(book)
```

The idea is that mypy should infer the proper types out of ormar typing so **WITHOUT ADDING ANY TYPE ANNOTATIONS** in `test_types.py`, the `reveal_type` should return proper types for all `reveal_type` calls. I managed to get queryset and model method working, some success with relations, what is dangling are reverse relations and querysetproxy. 

As everything with ormar since it moves between `pydantic`, `sqlalchemy` and `fastapi` to get the proper types it requires some type hacking 🤣 

In the so called meantime I made a mess with types and it will require cleaning but at least with some success:

The result of `mypy ormar tests` is as of now:

```
tests\test_types.py:76: note: Revealed type is 'tests.test_types.Author*'
tests\test_types.py:84: note: Revealed type is 'tests.test_types.Publisher*'
tests\test_types.py:85: note: Revealed type is 'ormar.relations.relation_proxy.RelationProxy[tests.test_types.Publisher*]'
tests\test_types.py:86: note: Revealed type is 'tests.test_types.Publisher*'
tests\test_types.py:89: note: Revealed type is 'Union[typing.Sequence[<nothing>], None]'  #WRONG
tests\test_types.py:91: note: Revealed type is 'Any'  #WRONG
tests\test_types.py:92: note: Revealed type is 'tests.test_types.Book*'
tests\test_types.py:93: note: Revealed type is 'builtins.list*[Union[tests.test_types.Book*, None]]'
tests\test_types.py:94: note: Revealed type is 'tests.test_types.Book*'
tests\test_types.py:95: note: Revealed type is 'ormar.queryset.queryset.QuerySet[tests.test_types.Book*]'
tests\test_types.py:96: note: Revealed type is 'tests.test_types.Author*'
tests\test_types.py:98: note: Revealed type is 'Any'  #WRONG
tests\test_types.py:100: note: Revealed type is 'Any'  #WRONG
tests\test_types.py:101: note: Revealed type is 'tests.test_types.Author*'
tests\test_types.py:102: note: Revealed type is 'builtins.str'
tests\test_types.py:103: note: Revealed type is 'Any'  #WRONG
tests\test_types.py:104: note: Revealed type is 'Any'  #WRONG
```

If you have time and you are willing to help you can check this branch and maybe you also have an idea how to fix rest of the types :) Cause for me it's like a try and error that's why I created a mess :D

dudil:
Hi @collerek - sorry for late response. If still relevant let me know and I'll be very much happy to see if I can assist!

Issue #137, opened 2021-03-24
[Feature request] Allow for composite primary keys, including foreign key columns as part of the PK


tillspeicher:
It would be quite useful to be able to use composite primary keys, i.e. primary keys consisting of multiple DB columns.
Supporting this feature would mean supporting composite foreign keys as well.
And ideally, in addition, composite primary keys should be able to also span foreign key columns.

Concretely, I would like to be able to define models as follows:
```python
class User(ormar.Model):
    id: uuid.UUID = ormar.UUID(primary_key=True)
    name: str = ormar.String(max_length=100)
    
# Note that this model has two primary key columns, one of which is a foreign key.
class UserItem(ormar.Model):
    user: User = ormar.ForeignKey(User, primary_key=True)
    id: int = ormar.Integer(primary_key=True)
    description: str = ormar.String(max_length=200)
    
# Implicitly this model has three primary key columns, because the user_item foreign
# key is a composite foreign key with both the User's id and the UserItem's id.
class UserItemStatus(ormar.Model):
    user_item: UserItem = ormar.ForeignKey(UserItem, primary_key=True)
    id: int = ormar.Integer(primary_key=True)
    status: str = ormar.String(max_length=100)
```

**Usecase:**
The reason for why I would like to use composite foreign keys is to combine the benefits of UUID and integer primary keys.
Using the example above, UUID PKs allow me to hide how many users are registered in my system, which would be visible with integer (auto-increment) keys.
But then, I would like to use integer keys for `UserItem` and `UserItemStatus` that increment on a *per-user basis*. I want to avoid using UUID keys for those items for reasons of column size and non-sequential data-layout, but also avoid a single integer primary key column because that would reveal how many items exist in the system when the ids are sent over an API.
Using integer keys that increment on a per-user basis is ideal here, but it requires to combine them with another column into the primary key, because they would not be unique otherwise.

Is this feature something you would like to support?
If yes, I could also try to implement this myself and open a PR.

SQLAlchemy has support for this feature: https://docs.sqlalchemy.org/en/14/core/constraints.html

An additional consideration:
The `UserItemStatus` model above has a foreign key relation `user_item`, but ideally it should also have a foreign key relation `user` since `user` is part of the `user_item` foreign key.
So there should be a way to define a `user` foreign key relationship that uses the column already present in the `user_item` relationship.
If you are interested in supporting this feature, I can think about how to define these two relationships in an efficient way.

collerek:
[...]

PR #138, opened 2021-03-27
Add initial set of tests for composite primary and foreign keys

tillspeicher:
I started this PR for the composite PK and FK implementation, please don't merge it yet.
So far, I only added a set of tests that covers the interfaces and behavior that we discussed in #137 . The tests likely contain errors, because there is no implementation of the actual functionality yet, so I couldn't run the code.
There are probably also some missing corner cases that should be tested as well.

For now, this is meant primarily as a specification for the composite PK and FK feature and to document the requirements.
In particular, the tests demonstrate and check composite PK- and FK-based CRUD operations, preventing direct access to FK columns, checking that reverse and many-to-many relationships are created correctly, that composite FKs are compatible with `pydantic` and that the `pk` property works as discussed.
In addition to what we discussed before, I added a test to ensure that composite PKs cannot use FK columns directly, as that would expose them via the `pk` property. So they need to use the names of the FK relations instead.

Please let me know if you would like anything to be changed or added @collerek . Otherwise I'll go ahead with the implementation.

### DIFF
--- a/tests/test_composite_keys.py
+++ b/tests/test_composite_keys.py
@@ -1,0 +1,424 @@
+from typing import (
+    List,
+    Optional,
+)
+import uuid
+
+import databases
+import pytest
+import sqlalchemy
+
+import ormar
+from tests.settings import DATABASE_URL
+
+database = databases.Database(DATABASE_URL, force_rollback=True)
+metadata = sqlalchemy.MetaData()
+
+
+class User(ormar.Model):
+    class Meta:
+        tablename = "users"
+        metadata = metadata
+        database = database
+
+    id: uuid.UUID = ormar.UUID(primary_key=True, default=uuid.uuid4)
+    email: str = ormar.String(nullable=False, max_length=100)
+
+
+class Project(ormar.Model):
+    class Meta:
+        tablename = "projects"
+        metadata = metadata
+        database = database
+        constraints = [
+            ormar.PrimaryKeyConstraint("id", "owner"),
+            ormar.ForeignKeyConstraint(User, ["owner_id"], ["id"], "owner"),
+        ]
+
+    id: int = ormar.Integer()
+    owner_id: uuid.UUID = ormar.UUID(default=uuid.uuid4)
+    name: str = ormar.String(nullable=False, max_length=100)
+
+
+class Tag(ormar.Model):
+    class Meta:
+        tablename = "tags"
+        metadata = metadata
+        database = database
+        constraints = [
+            ormar.PrimaryKeyConstraint("id", "owner", "project_id"),
+            ormar.ForeignKeyConstraint(
+                Project,
+                ["owner", "project_id"],
+                ["owner", "id"],
+                name="tag_project",
+            ),
+        ]
+
+    id: int = ormar.Integer(primary_key=True)
+    owner: User = ormar.ForeignKey(User)
+    project_id: int = ormar.Integer()
+    name: str = ormar.String(nullable=False, max_length=100)
+
+
+class TaskTag(ormar.Model):
+    class Meta:
+        tablename = "task_tags"
+        metadata = metadata
+        database = database
+
+    id: int = ormar.Integer(primary_key=True)
+    assigned_by: Optional[User] = ormar.ForeignKey(User)
+
+
+class Task(ormar.Model):
+    class Meta:
+        tablename = "tasks"
+        metadata = metadata
+        database = database
+        constraints = [
+            ormar.PrimaryKeyConstraint("id", "owner", "project"),
+            ormar.ForeignKeyConstraint(
+                User,
+                ["user_id"],
+                ["id"],
+                name="owner",
+                related_name="owners",
+                db_name="owner_fk",
+            ),
+            ormar.ForeignKeyConstraint(
+                Project,
+                ["owner_id", "project_id"],
+                ["owner", "id"],
+            ),
+        ]
+
+    id: int = ormar.Integer(primary_key=True)
+    owner_id: uuid.UUID = ormar.UUID()
+    project_id: int = ormar.Integer()
+    description: str = ormar.String(nullable=False, max_length=200)
+    completed: bool = ormar.Boolean(nullable=False, default=False)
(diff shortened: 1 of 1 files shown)


tests/test_composite_keys.py
```diff
@@ -0,0 +1,424 @@
...
+
+
+class TaskTag(ormar.Model):
+    class Meta:
+        tablename = "task_tags"
+        metadata = metadata
+        database = database
+
+    id: int = ormar.Integer(primary_key=True)
+    assigned_by: Optional[User] = ormar.ForeignKey(User)
```
collerek, line 71: Right now you cannot declare relationship in `Through` model -> that's one of the things I wanted to add gradually, but for now `ormar` creates the fields during `expanding_reverse_relations`, relations are **not** copied in inheritance, and explicitly stating any relation on through Model will raise exception. 

The key is that you can have multiple relations between two models, and `ormar` wouldn't know if the relation you created is the one that should be used for m2m relation or it should create other field for this relation (relation field name would have to be passed to `ManyToMany` probably).

collerek:
Ok, in general the tests look good, as you said those test are responsible for checks regarding Model declaration, especially getting/ setting pks. 

Since pk is such a crucial thing that is used almost everywhere (among others: order by, joins, filters, relations etc.), tests for this have to pretty much cover everything. Some of the following tests have common parts, but for clarity I listed what should be tested.

That includes, but is not limited to: 

*  All `Model` methods (`save/update/upsert/load` etc.)
*  All `QuerySet` methods (`filter/fields/get/delete/order_by` etc.)
*  All `QuerysetProxy` methods (mostly same as with `QuerySet`, separately for many to many and reverse side of fk as they are handled a little different)
*  Inheritance:
    * Inheritance from `Mixin` and `Concrete` models
    * PKs declared in parent, and redeclared in children
    * FKs declared in parent and redeclared with children
    * PKs and FKs declared only in children
    * Operations on inherited models for all `Model`, `QuerySet` and `QuerysetProxy`
* `Self-relations` and deferred resolving model with composite `pk` referencing itself
* Deferred resolving of `ForwardRefs` in relations (fks and m2m)
* Multiple relations between same two models 
    * two fks between A and B, 
    * two many to many between A and B, 
    * one fk and one m2m between A and B
* Optional and required FKs (now we didn't include a param for it,  probably `nullable: bool = True` should be included in FK signature)
* Columns of `pk` that are `autoincrement`/ have `default/default function` set and fields that are `required`. 
* Operations on one model alone
* Operations on simple relation between `A - B`
* Nested relations -> like `B - C - D - E`, with combination of fks and m2m
* Relations with same model appearing multiple times, schema similar to: 
  ```  
      B - C - Z
    /
  A               
    \  
      D - C - Y
  ```
*  Explicitly created and implicitly created `Through` Models for m2m relation (`Through` Models also can have composite pk? Probably yes.)
*  Complex filters (`and_/or_`)
* `dict()` and `json()` checks
*  Setting names for columns and relations:
    * aliases set for db columns in pks
    * aliases for db columns for fks
    * aliases for both pks and fks
    * default reverse relation names 
    * setting different reverse relation names
    * non existing column names used in composite pks and fks (exceptions)
    * overlapping related names (i.e. same `related_name` from two different models) causes exception
* Prohibiting encryption of pks and fks columns 
*  Internals:
    * Composite fields has to be stored in `Meta.model_fields` dictionary -> probably changing `ForeignKeyField` and `ManyToManyField` should be enough, don't thing a new field type would be required, maybe something like `is_composite` flag on relation fields.
    * Since pks cannot be partial the `pk` property for composites should return `None` (not empty dict) if not set and dict with values if set  
    * `ormar.Integer` fields that are part of composite `pks` should be set to `autoincrement=True` by default
    *  pks are always included in the model, even if someone explicitly `exclude_fields({pk_column(s)})` this call is "ignored" (pk column is added at the end regardless of anything else).
    * don't forget to `black/flake8` and `mypy` the codebase -> for flake8 only `ormar` is important not the tests, for `mypy/black` both
 
Maybe something else will come to my mind. As you can see there is a vast amount of areas that needs to be checked to ensure proper working of `ormar` with this new feature. I can probably help you with writing those tests if you want me to ;)

tillspeicher:
Thanks for taking the time to go over the tests and creating the list of requirements!
Indeed, those are a lot of tests :) It feels to me like the list you mentioned covers a significant part of the general Ormar functionality. In that case, it might be easier to cover those cases by modifying the existing set of tests by changing or adding some models that have composite PKs and FKs.
For now, I want to start with the actual implementation, but I will look into modifying the tests to use composite PKs and FKs once that's done and you are fine with modifying the existing tests rather than writing additional ones. Of course, any help in doing that would be welcome :)

Mypy on the tests is producing errors for now, but that's expected since e.g. `ForeignKeyConstraint` is not implemented yet.
However, when I ran black as described in the contribution guidelines on the codebase, it reformatted a number of ormar and test files for me. As far as I can see those modifications were done to reduce the line length. Are you using a max-line-length that's different from the 88 characters default?

collerek:
No, the lines length should be 88 if I remember correctly (that's also my pycharm and flake8 setting).

Yeah, that's the problem with pks -> they are used all over the place -> relations need to be resolved to them in joins, it's a default sort order if you do not provide additional one, they cannot be excluded in fields/exclude_fields etc. that's why a lot of tests are needed with this functionality to properly check all aspects on all db backends.

I'm ok with changing some of the tests (or preferably adding new models), but not all since then you loose tests for simple one column pks, and like they say never too much tests. I will help you with tests as that may be also an opportunity to clean them up a little, maybe divide into subpackages as it's starting to be a lot of files to grasp.

collerek:
@tillspeicher are you gonna continue this work or should I close this PR?

tillspeicher:
I'm sorry for going silent on the PR.
I got sidetracked with a different project than the one I need this for and had to prioritize that.
I have composite PKs implemented into the existing codebase with 15 tests still failing but haven't implemented an API for defining composite PKs and FKs yet, which will take some more time. I also need to merge this with the changes that happened to the codebase in the meantime.
I am planning on getting back to this issue, but can't say when yet. It's probably best to close the PR for now, once I get back to this project I can start a new one or reopen it.

PR #173, opened 2021-04-22
Fix typo in docstring

johnthagen:


johnthagen:
@collerek Somewhat related, why does `all()` return `List[Optional["T"]]` when the docs say:

```
If there are no rows meeting the criteria an empty list is returned.
```

Shouldn't that indicate that it returns `List["T"]` and it will just be an empty `[]` if it doesn't find anything? What does `None` communicate?

The `Optional` seems to flow form:

```py
    def _process_query_result_rows(self, rows: List) -> List[Optional["T"]]:
        """
        Process database rows and initialize ormar Model from each of the rows.

        :param rows: list of database rows from query result
        :type rows: List[sqlalchemy.engine.result.RowProxy]
        :return: list of models
        :rtype: List[Model]
        """
        result_rows = [
            self.model.from_row(
                row=row,
                select_related=self._select_related,
                excludable=self._excludable,
                source_model=self.model,
                proxy_source_model=self.proxy_source_model,
            )
            for row in rows
        ]
        if result_rows:
            return self.model.merge_instances_list(result_rows)  # type: ignore
        return cast(List[Optional["T"]], result_rows)
```

collerek (commented):
Thanks for catching that, and it should be `(*args, **kwargs)` actually as `args` refer to filter groups

collerek:
As for `Optional` you are right, not sure why it's there maybe it's coming from old code or maybe it was always a bug. If you can, please remove optional and make sure `ormar` checks with `mypy` (so whole method chain needs to be adjusted probably).

collerek:
Come to think about it i think i forgot to update the do docstrings for all queryset and queryset proxy methods, i just updated the docs. All of them, if accept kwargs and args, and refer to filter in docstrings should have `filter(*args, **kwargs)`

johnthagen:
> As for Optional you are right, not sure why it's there maybe it's coming from old code or maybe it was always a bug. If you can, please remove optional and make sure ormar checks with mypy (so whole method chain needs to be adjusted probably).

I broke this into its own issue, #173, so we don't forget about it as I can't get to it today.

johnthagen:
> Come to think about it i think i forgot to update the docstrings for all queryset and queryset proxy methods, i just updated the docs 🙈. All of them, if accept kwargs and args, and refer to filter in docstrings should have filter(*args, **kwargs)

@collerek Based on this I'm not sure what needs to be changed in this PR. Should it be closed?

Or did you mean the change should be:

```
Passing kwargs is a shortcut and equals to calling `filter(*args, **kwargs).all()`.
```

collerek:
> Or did you mean the change should be:
> 
> ```
> Passing kwargs is a shortcut and equals to calling `filter(*args, **kwargs).all()`.
> ```

The second one, so: 
```python
# Passing args and/or kwargs is a shortcut and equals to calling `filter(*args, **kwargs).all()`.
```

And the same applies for other methods: `get`, `get_or_none`, `get_or_create` etc. 
All methods from `QuerySet` that now accepts also `*args` and has this reference to `filter()` in docstring should be updated like the line above. The same applies to equivalent methods located in `Querysetproxy`.

If you can update them all I would be grateful, if not I will do this in next release.

Issue #285, opened 2021-07-24
Provide a way to customize the model Config class


naturalethic:
**Is your feature request related to a problem? Please describe.**
I would like to customize the model config, specifically to specify `allow_mutation = False`.

**Describe the solution you'd like**
If a `Config` class is defined in the model, use it instead of `get_pydantic_base_orm_config`/

**Describe alternatives you've considered**
The only solution I've found so far is to subclass ormar's ModelMetaclass and override the __new__ method.

collerek:
Yeah I had it on mind for quite some time.

You still can edit the config after class declaration as it's stored in `__config__` attribute, so after your class declaration you can set:
```python
class Model(ormar.Model):
    ...

Model.__config__.allow_mutation = False
```

But allowing to overwrite the config in class declaration should be the preferred way.

naturalethic:
Oh great, that will work for me for now, thanks.  I was just typing up a patch, so I'll keep that below, I'm sure you're aware of this possible solution.

In `ormar.models.metaclass` around line `560`, replace:
```
        attrs["Config"] = get_pydantic_base_orm_config()
```
with:
```
        if not "Config" in attrs:
            attrs["Config"] = get_pydantic_base_orm_config()
```

collerek:
Yep, if you can issue a PR with this solution that would be welcomed too! :)

naturalethic:
https://github.com/collerek/ormar/pull/286

collerek:
Released in 0.10.16 - thanks again! :)

PR #286, opened 2021-07-24
Allow custom model config

naturalethic:
https://github.com/collerek/ormar/issues/285

The approach I took is to use a mixin.


collerek:
Thanks for the PR! :)

naturalethic:
This did fail the mypy tests.  I'm not familiar with mypy.

PR #300, opened 2021-08-09
Correct typo

sondrelg:
I was just skimming the docs (which are great 🎉) and spotted what I believe is a tiny typo 🙂


collerek:
Thanks for catching that and the fix☺️

And thank you for kind words, that means a lot😁

Issue #337, opened 2021-09-06
Model.dict() is taking to much time to respond


luisjimenez6245:
**Describe the bug**
Im using a model with 1 many to many rel and 1 foreign key, with ormar, but the seralization is taking too much to do it. After disabling dict() the overloaded method, the response model does not have any related model, but it takes almost a 1/100  (I didn't make the exact calculations) of the time to do it. This problem also affects  to make the json response, so we should expect a slow response time. Is there any way to disable this override method or is there any plan/idea to optimize it? 

With ormar dict():
<img width="453" alt="Captura de Pantalla 2021-09-06 a la(s) 12 29 23" src="https://user-images.githubusercontent.com/26035464/132249156-d4553924-3c82-48c9-b57b-b1489004538e.png">

With only-pydantic dict():
<img width="410" alt="Captura de Pantalla 2021-09-06 a la(s) 12 28 13" src="https://user-images.githubusercontent.com/26035464/132249158-708ca324-e146-4fa8-8f6e-5f735cdaf969.png">

The list of models that I´m loading doesn't have any related model loaded, neither items on the many to many relation

collerek:
I'm trying to optimize the `dict()` if I can on a regular basis but to check your scenario I would need your models and sample data to make some tests.

That seems like a quite long response indeed, do you load the related models in the query or not?

Because if you do, then you serialize the main model plus n x related model, where n is a number of related model instances of the main model. So you are comparing the time of serializing one model (pydantic) versus n+1 models (ormar).

Do you want the related model included in the dict or not?

luisjimenez6245:
In answer to your questions, im not loading the related models.

Here are my principal models that im working on:


```python
class Release(orm.Model):

    id: int = orm.Integer(name="id", primary_key=True)
    title: str = orm.String(max_length=200, default='')
    description: str = orm.Text(default='')
    tag: Tag = orm.ForeignKey(Tag, ondelete='CASCADE', onupdate='CASCADE')
    changelogs:  List[Changelog] = orm.ManyToMany(
        Changelog, through=ChagenlogRelease, ondelete='CASCADE', onupdate='CASCADE'
    )
    data: dict = orm.JSON(default={})

class Project(orm.Model):

    id: int = orm.Integer(name="id", primary_key=True)
    name: str = orm.String(max_length=100)
    description: str = orm.Text(default='')
    git_url: str = orm.String(max_length=500, default='')
[...]

PR #338, opened 2021-09-06
add __relation_map__ property to only calculate one time the relation_map

luisjimenez6245:
#337


luisjimenez6245:
it will fail if something has a different __relation_map__

PR #342, opened 2021-09-10
Add recursion limit #337

luisjimenez6245:


luisjimenez6245:
@collerek what do you think about making this change to solve the number of recursion calls that im making? #337

PR #364, opened 2021-10-05
Fix grammar in queries docs

johnthagen:
I believe `completeness` was the intended word rather than `complicity`?


collerek:
Ha ha, good catch - that's what you get from autocompletion 😆 
Thanks again! ;)

PR #366, opened 2021-10-06
Add tip about using values() to exclude required fields

johnthagen:
Related to https://github.com/collerek/ormar/discussions/363


collerek:
Thanks 🙏

PR #429, opened 2021-11-15
Fix populate_default_values

Kurbezz:
Bug:  Set field value to null when specifying server_default_value and bool (value) == False (empty string, False or 0).

```
class ExampleModel(Model):
  some_field = Integer(server_default_value=sa.text('0'), default=0, nullable=False)

await ExampleModel.objects.create(some_field=0)
```


collerek:
Awesome, thank you! 😁

Can you add a test (in test model definitions) for this case so it won't happen again?

collerek:
Thanks a lot! 😊

PR #493, opened 2021-12-21
 use the dict to store the receivers in signal

ponytailer:
- use the dict to store the receivers in signal  
- SignalEmitter extends dict to save the signals

 @collerek

### DIFF
--- a/docs/fields/field-types.md
+++ b/docs/fields/field-types.md
@@ -140,13 +140,6 @@ You can pass `timezone=True` for timezone aware database column.
 * Sqlalchemy column: `sqlalchemy.JSON`  
 * Type (used for pydantic): `pydantic.Json` 
 
-### JSONB
-
-`JSONB()` has no required parameters, only support it in the postgresql.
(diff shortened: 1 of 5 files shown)


ponytailer:
The `Signal` is simple and useful. it should be common, should move the `sender` into the `kwargs` maybe ?

collerek:
I avoid adding dialect-specific fields to `ormar`, `JSONB` is already supported in an awesome package: https://github.com/tophat/ormar-postgres-extensions by @etimberg. 

I will check the dict signals, don't remember why it was a list.

ponytailer:
> https://github.com/tophat/ormar-postgres-extensions

Fine, I remove the `jsonb`, maybe should support here, not the extra lib.

ponytailer:
> The `Signal` is simple and useful. it should be common, should move the `sender` into the `kwargs` maybe ? @collerek

And this ?

collerek:
Awesome, thanks!

PR #520, opened 2022-01-11
improve bulk_create

Mng-dev-ai:
Use execute instead of execute_many for better performance as database.execute_many is actually running execute in a loop in databases.


collerek:
Hmm, I think ponytailer changed this line at the same time and I merged that PR first, and you need to actually add/change some line instead of only removing one.

If you would like to be included, can you push another commit adding the comment:

```python
# don't use execute_many, as in databases it's executed in a loop 
# instead of using execute_many from drivers
expr = self.table.insert().values(ready_objects)
``` 

And I will merge it again. 
Credit in release notes goes to you anyway :)

collerek:
Sorry about that 😢

collerek:
Let me know if you want to push this commit or not before the release that is ready

Mng-dev-ai:
Yes, Sure I want to push this commit.

Mng-dev-ai:
So, Should I make a new merge request right ?

Mng-dev-ai:
Done at https://github.com/collerek/ormar/pull/525#issue-1103952724.

PR #524, opened 2022-01-14
add post-signal in bulk-update & bulk-create

ponytailer:
- fix typo
- add post-signal in bulk-update and bulk-create

### DIFF
--- a/ormar/models/mixins/save_mixin.py
+++ b/ormar/models/mixins/save_mixin.py
@@ -57,6 +57,21 @@ class SavePrepareMixin(RelationMixin, AliasMixin):
         return new_kwargs
 
     @classmethod
+    def prepare_model_to_update(cls, new_kwargs: dict) -> dict:
+        """
+        Combines all preparation methods before updating.
+        :param new_kwargs: dictionary of model that is about to be saved
+        :type new_kwargs: Dict[str, str]
+        :return: dictionary of model that is about to be updated
+        :rtype: Dict[str, str]
+        """
+        new_kwargs = cls.parse_non_db_fields(new_kwargs)
+        new_kwargs = cls.substitute_models_with_pks(new_kwargs)
+        new_kwargs = cls.reconvert_str_to_bytes(new_kwargs)
+        new_kwargs = cls.translate_columns_to_aliases(new_kwargs)
+        return new_kwargs
+
+    @classmethod
     def _remove_not_ormar_fields(cls, new_kwargs: dict) -> dict:
         """
         Removes primary key for if it's nullable or autoincrement pk field,
--- a/ormar/queryset/queryset.py
+++ b/ormar/queryset/queryset.py
@@ -1,3 +1,4 @@
+import asyncio
 from typing import (
     Any,
     Dict,
@@ -1049,17 +1050,23 @@ class QuerySet(Generic[T]):
         :param objects: list of ormar models already initialized and ready to save.
         :type objects: List[Model]
         """
-        ready_objects = []
-        for objt in objects:
-            new_kwargs = objt.dict()
-            new_kwargs = objt.prepare_model_to_save(new_kwargs)
-            ready_objects.append(new_kwargs)
+        ready_objects = [
+            obj.prepare_model_to_save(obj.dict())
+            for obj in objects
+        ]
+        expr = self.table.insert().values(ready_objects)
+        # shouldn't use the execute_many, it's `queries.foreach(execute)`
+        await self.database.execute(expr)
 
-        expr = self.table.insert()
-        await self.database.execute_many(expr, ready_objects)
+        # FIXME: add pre_save signals
+        async def after_create(entity):
+            entity.set_save_status(True)
+            await entity.signals.post_save.send(
+                sender=entity.__class__, instance=entity)
 
-        for objt in objects:
-            objt.set_save_status(True)
+        await asyncio.gather(
+            *[after_create(entity) for entity in objects]
+        )
 
     async def bulk_update(  # noqa:  CCR001
         self, objects: List["T"], columns: List[str] = None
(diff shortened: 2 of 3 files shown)


ponytailer:
Sorry,  I saw the part of changes including #520 just now

collerek:
Yes, I agree that if implemented those should be separate signals that accept a list of instances -> that way you can skip the bool flag from method signatures and still have some signals only for "normal" create or update but not for bulk operations.

But there is a bigger problem and a reason why this is not present. 
I intentionally didn't add signals to bulk operations because bulk operations in any of the supported drivers do not return the updated data. So marking those objects as saved in your code is not really true, as you won't have primary key values for autoincrement fields, or won't have updated columns that have defaults set in the database.

And since they don't have pks (or at least don't have to have them) there is no reliable way to refresh the models after bulk_create, so the signal in your code receives incomplete data (each instance in db differs from the one you have in bulk operation).

ponytailer:
Thanks for your review.  Please release when it was merged.

I need this signal to develop project.

ormar/queryset/queryset.py
```diff
@@ -1049,25 +1053,25 @@ async def bulk_create(self, objects: List["T"]) -> None:
...
 
-        expr = self.table.insert()
-        await self.database.execute_many(expr, ready_objects)
+        # shouldn't use the execute_many, it's `queries.foreach(execute)`
+        await self.database.execute(expr)
 
-        for objt in objects:
-            objt.set_save_status(True)
+        for obj in objects:
+            obj.set_save_status(True)
```
collerek, line 1066: When now I think about this, this is not true, as objects don't have database side populated columns refreshed (so you cannot use this object as a factual representation of a db state). At the same time it actually was saved, but not fetched, need to think about this.

ormar/queryset/queryset.py
```diff
@@ -1076,11 +1080,14 @@ async def bulk_update(  # noqa:  CCR001
 
         Bulk operations do not send signals.
 
-        :param objects: list of ormar models
+        :param objects:list of ormar models
```
collerek, line 1083: Missing space

ormar/queryset/queryset.py
```diff
@@ -1127,5 +1132,11 @@ async def bulk_update(  # noqa:  CCR001
...
         await self.database.execute_many(expr, ready_objects)
 
-        for objt in objects:
-            objt.set_save_status(True)
+        for obj in objects:
+            obj.set_save_status(True)
+
+        entity = list(objects)[0]
+        await entity.signals.post_bulk_update.send(
+            sender=entity.__class__, instances=objects
```
collerek, line 1140: Use `self.__class__` instead of extracting the entity from a list. 

Since as of now `ormar` does not support onupdate database triggers there is no need for refresh from db, so should be good to go.

ormar/queryset/queryset.py
```diff
@@ -1076,11 +1080,14 @@ async def bulk_update(  # noqa:  CCR001
...
         Bulk operations do not send signals.
 
-        :param objects: list of ormar models
+        :param objects:list of ormar models
         :type objects: List[Model]
         :param columns: list of columns to update
         :type columns: List[str]
         """
+        if not objects:
+            raise ModelListEmptyError("Bulk update objects are empty!")
```
collerek, line 1089: This one is kind of a breaking change, but I can agree that raising an explicit exception on an empty list makes sense and shouldn't actually break much.

ormar/models/mixins/save_mixin.py
```diff
@@ -56,6 +56,21 @@ def prepare_model_to_save(cls, new_kwargs: dict) -> dict:
         new_kwargs = cls.translate_columns_to_aliases(new_kwargs)
         return new_kwargs
 
+    @classmethod
+    def prepare_model_to_update(cls, new_kwargs: dict) -> dict:
```
collerek, line 60: I thought about this for some time already 👍

ormar/queryset/queryset.py
ponytailer: maybe should support

ormar.datetime(onupdate=datetime.now)

ormar/queryset/queryset.py
ponytailer: maybe should support

ormar.datetime(onupdate=datetime.now)

collerek: That can be considered in other PR

ponytailer: yes.  I think it will be supported later.

ormar/queryset/queryset.py
ponytailer:
```
if not objects:
    logger.warning('bulk update objects are empty')
    return
```

And,  log the empty and return here only?

ormar/queryset/queryset.py
```diff
@@ -1127,5 +1132,10 @@ async def bulk_update(  # noqa:  CCR001
         expr = str(expr)
         await self.database.execute_many(expr, ready_objects)
 
-        for objt in objects:
-            objt.set_save_status(True)
+        for obj in objects:
+            obj.set_save_status(True)
+
+        await self.__class__.signals.post_bulk_update.send(
```
ponytailer, line 1138: How to get the signals here?  @collerek

ormar/queryset/queryset.py
collerek: Sorry it should be `self.model_cls`, forgot we are in `QuerySet` not `Model`.

ormar/queryset/queryset.py
collerek: You can be explicit and go with `self.model_cls.Meta.signals`, don't remember if it's a property of model or class attribute

collerek:
```
@property
    def signals(self) -> "SignalEmitter":
        """Exposes signals from model Meta"""
        return self.Meta.signals
```

Ok, thats a property of model, so you will get descriptor back from `self.model_cls.signals` and need to go with `self.model_cls.Meta.signals`. Sorry, didn't check before.

ormar/queryset/queryset.py
```diff
@@ -1127,5 +1132,10 @@ async def bulk_update(  # noqa:  CCR001
         expr = str(expr)
         await self.database.execute_many(expr, ready_objects)
 
-        for objt in objects:
-            objt.set_save_status(True)
+        for obj in objects:
+            obj.set_save_status(True)
+
+        await obj.signals.post_bulk_update.send(
+            sender=self.__class__, instances=objects
```
collerek, line 1139:
```python
await self.model_cls.Meta.signals.post_bulk_update.send(
            sender=self.model_cls, instances=objects
)
```

PR #532, opened 2022-01-17
add onupdate option in field simply

ponytailer:
- support option `onupdate` in field simply

### DIFF
--- a/ormar/fields/base.py
+++ b/ormar/fields/base.py
@@ -1,5 +1,6 @@
 import warnings
-from typing import Any, Dict, List, Optional, TYPE_CHECKING, Type, Union
+from typing import Any, Callable, Dict, List, Optional, TYPE_CHECKING, Type, \
+    Union
 
 import sqlalchemy
 from pydantic import Json, typing
@@ -47,6 +48,8 @@ class BaseField(FieldInfo):
         self.index: bool = kwargs.pop("index", False)
         self.unique: bool = kwargs.pop("unique", False)
         self.pydantic_only: bool = kwargs.pop("pydantic_only", False)
+        self.onupdate: Union[Callable[..., Any], Any, None] = kwargs.pop(
+            "onupdate", None)
         if self.pydantic_only:
             warnings.warn(
                 "Parameter `pydantic_only` is deprecated and will "
@@ -216,6 +219,33 @@ class BaseField(FieldInfo):
         return self.ormar_default is not None or (
             self.server_default is not None and use_server
         )
+
+    def has_onupdate(self) -> bool:
+        """
+        Checks if the field has onupdate value set.
+        :return: result of the check if onupdate value is set
+        rtype: bool
+        """
+        if self.__pydantic_type__ is None:
+            return self.onupdate is not None
+
+        if self.onupdate is not None and not callable(self.onupdate):
+            if isinstance(self.onupdate, self.__pydantic_type__):
+                return True
+        elif self.onupdate is not None and callable(self.onupdate):
+            return True
+        return False
+
+    def get_onupdate(self) -> Union[None, Any]:
+        """
+        Get onupdate value if set
+
+        :return: result of the onupdate
+        rtype: Any
+        """
+        if callable(self.onupdate):
+            return self.onupdate()
+        return self.onupdate
 
     def is_auto_primary_key(self) -> bool:
         """
--- a/ormar/models/mixins/save_mixin.py
+++ b/ormar/models/mixins/save_mixin.py
@@ -77,6 +77,7 @@ class SavePrepareMixin(RelationMixin, AliasMixin):
         new_kwargs = cls.substitute_models_with_pks(new_kwargs)
         new_kwargs = cls.reconvert_str_to_bytes(new_kwargs)
         new_kwargs = cls.dump_all_json_fields_to_str(new_kwargs)
+        new_kwargs = cls.populate_onupdate_value(new_kwargs)
         new_kwargs = cls.translate_columns_to_aliases(new_kwargs)
         return new_kwargs
 
@@ -236,6 +237,21 @@ class SavePrepareMixin(RelationMixin, AliasMixin):
                 and new_kwargs.get(field_name, None) is None
             ):
                 new_kwargs.pop(field_name, None)
+        return new_kwargs
+
+    @classmethod
+    def populate_onupdate_value(cls, new_kwargs: Dict) -> Dict:
+        """
+        Populate value which from onupdate options in field
+
+        :param new_kwargs: dictionary of model that is about to be saved
+        :type new_kwargs: Dict
+        :return: dictionary of model that is about to be saved
+        :rtype: Dict
+        """
+        for field_name, field in cls.Meta.model_fields.items():
+            if field.has_onupdate() and not field.pydantic_only:
+                new_kwargs[field_name] = field.get_onupdate()
         return new_kwargs
 
     @classmethod
--- a/ormar/models/model.py
+++ b/ormar/models/model.py
@@ -1,4 +1,5 @@
-from typing import Any, Dict, List, Optional, Set, TYPE_CHECKING, TypeVar, Union
+from typing import Any, Dict, List, Optional, Set, TYPE_CHECKING, TypeVar, \
+    Union
 
 import ormar.queryset  # noqa I100
 from ormar.exceptions import ModelPersistenceError, NoMatch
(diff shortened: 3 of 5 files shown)


ormar/fields/base.py
```diff
@@ -217,6 +220,33 @@ def has_default(self, use_server: bool = True) -> bool:
...
+    def get_onupdate(self) -> Union[None, Any]:
+        """
+        Get onupdate value if set
+
+        :return: result of the onupdate
+        rtype: Any
+        """
+        if callable(self.onupdate):
+            return self.onupdate()
+        return self.onupdate
```
collerek, line 248: This can be simplifed to:
```python
return self.onupdate() if callable(self.onupdate) else self.onupdate
```

ormar/models/mixins/save_mixin.py
```diff
@@ -402,3 +418,11 @@ def _get_field_values(self, name: str) -> List:
         if not isinstance(values, list):
             values = [values]
         return values
+
+    @classmethod
+    def get_fields_has_onupdate(cls) -> List[str]:
```
collerek, line 423: Would rename this to `get_fields_with_onupdate`.

Also, when it's called the first time model should be fully populated (relations etc.) so we can cache those fields in a property `_on_update_fields` cause it will be run with each `update`/`bulk_update` and return from _cache if this attribute is not none to skip iterating through all fields every time. Something similar to `cls.extract_related_names()`

ormar/models/mixins/save_mixin.py
```diff
@@ -238,6 +239,21 @@ def populate_default_values(cls, new_kwargs: Dict) -> Dict:
...
+
+        :param new_kwargs: dictionary of model that is about to be saved
+        :type new_kwargs: Dict
+        :return: dictionary of model that is about to be saved
+        :rtype: Dict
+        """
+        for field_name, field in cls.Meta.model_fields.items():
+            if field.has_onupdate() and not field.pydantic_only:
+                new_kwargs[field_name] = field.get_onupdate()
+        return new_kwargs
```
collerek, line 255: This should use the `get_fields_with_onupdate` defined below to remove repetition and utilize the cache

```python
for field_name in cls.get_fields_with_onupdate():
    new_kwargs[field_name] = cle.Meta.model_fields[field_name].get_onupdate()
return new_kwargs
```

Also, this does not allow explicit overwrites (if you want to manually overwrite the field, it will be changed to onupdate value anyway).

ponytailer:
Cascade should be another option in field ?  @collerek

collerek:
> Cascade should be another option in field ? @collerek
> 
> The `onupdate` don't handle the relations.
> 
> such like the sqlalchemy, and the `cascade` should define in the database.
> 
> ```
> book = ormar.ForiegnKey(Book, cascade=CASCADE.DELETE)
> ```

Yes, it does.
There are already onupdate and ondelete, but only for foreign keys.

https://github.com/collerek/ormar/blob/8dc05d5cf48469f6285e88ab87d38b57c678976f/ormar/fields/foreign_key.py#L185-L196

It's used to populate onupdate and ondelete foreign key options in sqlalchemy foreign key:

https://github.com/collerek/ormar/blob/8dc05d5cf48469f6285e88ab87d38b57c678976f/ormar/fields/base.py#L242-L252

Where you pass a string according to sqlalchemy core specification:
https://docs.sqlalchemy.org/en/14/core/constraints.html#on-update-and-on-delete

ormar/models/mixins/save_mixin.py
```diff
@@ -238,6 +239,22 @@ def populate_default_values(cls, new_kwargs: Dict) -> Dict:
...
+        Populate value which from onupdate options in field
+
+        :param new_kwargs: dictionary of model that is about to be saved
+        :type new_kwargs: Dict
+        :return: dictionary of model that is about to be saved
+        :rtype: Dict
+        """
+        for field_name in cls.get_fields_with_onupdate():
+            field = cls.Meta.model_fields[field_name]
+            if field.has_onupdate() and not field.pydantic_only:
```
collerek, line 254: ifs are redundant, you already checked it in get_fields_with_onupdate

ponytailer:
> 
That's in foreignKey only. 

when the model was updated, something to do auto, such as the `updated_time`

otherwise, I have to define the code
```

async def update(self,  **kwargs):
     kwargs["updated_time"] = get_ts()
     super().update(**kwargs)

```

ormar/models/mixins/save_mixin.py
```diff
@@ -402,3 +419,11 @@ def _get_field_values(self, name: str) -> List:
         if not isinstance(values, list):
             values = [values]
         return values
+
+    @classmethod
+    def get_fields_with_onupdate(cls) -> List[str]:
+        return [
```
collerek, line 425: You didn't add a cached property like i.e. `_related_fields` in:

https://github.com/collerek/ormar/blob/8dc05d5cf48469f6285e88ab87d38b57c678976f/ormar/models/mixins/relation_mixin.py#L36-L53

ormar/fields/base.py
```diff
@@ -47,6 +48,8 @@ def __init__(self, **kwargs: Any) -> None:
         self.index: bool = kwargs.pop("index", False)
         self.unique: bool = kwargs.pop("unique", False)
         self.pydantic_only: bool = kwargs.pop("pydantic_only", False)
+        self.onupdate: Union[Callable[..., Any], Any, None] = kwargs.pop(
```
collerek, line 51: I understand what you want and I agree - that's a nice feature and thanks. ;)

What I'm trying to say is that `ForeignKeyField` inherits from `BaseField` and defines already `onupdate` field.
So now we have one attribute in fields that is used either for FK cascades (already implemented), or for updates of the values during updates (this PR).

Simply change how the field is stored here:
```
self.on_update: Union[Callable[..., Any], Any, None] = kwargs.pop(...) # use on_update instead of onupdate name
``` 

That way we do not overwrite this in FK as this is a different functionality.

ormar/models/model.py
```diff
@@ -220,6 +223,7 @@ async def update(self: T, _columns: List[str] = None, **kwargs: Any) -> T:
         :rtype: Model
         """
         if kwargs:
+            kwargs = self.populate_onupdate_value(kwargs)
```
collerek, line 226: That means we always overwrite the field value with onupdate value.

If someone specifies the value for a column explicitly in an update it should have priority over the onupdate value (so you can explicitly overwrite it if you want).

You expressed the same idea in docs above where you wrote: `if you don't update the field which has the onupdate option, its value will be changed from onupdate definition`. But with the current implementation, it's always overwritten, even if you do update the field explicitly.

ponytailer:
This is pr only include the `update`,   not the bulk-update.
because
```
        for obj in objects:
            # when the obj.__setattr__, should be dirty for column
            # only load the kv from dirty fields, not all.
            new_kwargs = obj.dict()
```

collerek:
Can you restore/uncomment `bulk_update` changes? 
For now let's keep the changes logic as you proposed, without keeping status of each field (dirty or not)

ponytailer:
> Can you restore/uncomment `bulk_update` changes? 
> 
> For now let's keep the changes logic as you proposed, without keeping status of each field (dirty or not)
> 
> 
sorry. I'm on vacation now. Maybe I can create the new pr for bulk-update in a few days.

collerek:
> > Can you restore/uncomment `bulk_update` changes?
> > For now let's keep the changes logic as you proposed, without keeping status of each field (dirty or not)
> 
> sorry. I'm on vacation now. Maybe I can create the new pr for bulk-update in a few days.

Ahh of course. Happy Lunar New Year! 🥳 🎉 

No worries it can wait for your come back :)

ponytailer:
> > > Can you restore/uncomment `bulk_update` changes?
> > > For now let's keep the changes logic as you proposed, without keeping status of each field (dirty or not)
> > 
> > 
> > sorry. I'm on vacation now. Maybe I can create the new pr for bulk-update in a few days.
> 
> Ahh of course. Happy Lunar New Year! 🥳 🎉
> 
> No worries it can wait for your come back :)

Next pr will define the `dirty-columns` to update model dynamically, when the `columns` is None

collerek:
Before you implement something describe what you want to do. 
I told you that keeping the dirty status on fields might have a big performance influence and can be unreliable.

ponytailer:
> Before you implement something describe what you want to do. I told you that keeping the dirty status on fields might have a big performance influence and can be unreliable.

Thanks for your advice !

ormar/queryset/queryset.py
```diff
@@ -1111,7 +1114,15 @@ async def bulk_update(  # noqa:  CCR001
 
         columns = [self.model.get_column_alias(k) for k in columns]
 
+        # on_update_fields = {
```
collerek, line 1117: please either update the code or remove it -> we don't want commented out code in the code base

vvanglro:
Hi guys why is this PR not merged, I wish there was an onupdate like sqlalchemy.

Issue #386, opened 2021-10-14
`get_or_create` to return an additional Boolean field


amalshaji:
**Is your feature request related to a problem? Please describe.**

The `get_or_create` currently returns a `Model`. Add a boolean(`[Model, bool]`) to specify whether the item was created or not. Similar to [django get_or_create](https://docs.djangoproject.com/en/3.2/ref/models/querysets/#get-or-create)

**Describe the solution you'd like**

By going through the [ormar get_or_create](https://github.com/collerek/ormar/blob/7f0eee2936aa24811a38a129619170ab3420ce81/ormar/queryset/queryset.py#L952) code, the function can implement it by returning a bool. Modified body:

```python
    ...
    try:
        return await self.get(*args, **kwargs), False
    except NoMatch:
        return await self.create(**kwargs), True
```

Any plans to add this feature? Would you accept a PR?

collerek:
Added in 0.11.0 - please update and check.

PR #537, opened 2022-01-18
add method `find_or_create`

ponytailer:
Should be define the new method to solve #386 , until the next big release to publish and remove this method.

BTW in discussions,  the QA:

How to define the `queryset` like that.

```
db.session.query(A,B).filter(A.aid == B.bid)
```


ponytailer:
maybe unnecessary.

PR #538, opened 2022-01-18
Bugfix: dumps the object which has the json fields in bulk create

ponytailer:
- bugfix for  `orjson.dumps -> bytes,  but expected str`
- define the `__queryset_cls__`

BTW,  `merge and squash` may be better.


collerek:
What is the bug that this PR fixes?

ponytailer:
> What is the bug that this PR fixes?

when use the bulk_create and object has the json field, 

`save_mixins.dump_all_json_fields_to_str: orjson.dumps return the bytes, not the str`

tests/test_queries/test_utils.py
```diff
@@ -0,0 +1,15 @@
+import json
```
collerek, line 1: This doesn't test orjson

ormar/queryset/utils.py
```diff
@@ -17,6 +17,15 @@
     from ormar import Model, BaseField
 
 
+def to_str(val: Union[bytes, str, int]) -> str:
```
collerek, line 20: There already is:

https://github.com/collerek/ormar/blob/8dc05d5cf48469f6285e88ab87d38b57c678976f/ormar/fields/parsers.py#L40-L43

which probably can be re-used

tests/test_queries/test_queryset_level_methods.py
```diff
@@ -76,6 +78,27 @@ class Meta:
...
+    class QuerySetCls(QuerySet):
+
+        async def first_or_404(self, *args, **kwargs):
+            entity = await self.get_or_none(*args, **kwargs)
+            if not entity:
+                # maybe HTTPException in fastapi
+                raise ValueError("customer not found")
+            return entity
+
+    __queryset_cls__ = QuerySetCls
```
collerek, line 96: This should be added to docs in sections of models as a way to overwrite setting for main/default `QuerySet`

ormar/queryset/utils.py
ponytailer: All use the `to_str`.

collerek (approved):

collerek:
Thanks for catching the bug and this PR!

I refactored to avoid code duplication.

Also, all model configs should go into `model.Meta` so I moved the `queryset_class` config parameter.

ponytailer:
> Thanks for catching the bug and this PR!
> 
> I refactored to avoid code duplication.
> 
> Also, all model configs should go into `model.Meta` so I moved the `queryset_class` config parameter.

ok, got it.

Release the version for the bugfix, please.

Issue #554, opened 2022-01-31
enhancement in get_or_create


MojixCoder:
In Django, we have `get_or_create(defaults=None, **kwargs)`.
It helps when we want to get a object with a specific field but if the object does not exist, create it with some additional values.   

This is how it's done in Django:
```python
obj, created = Person.objects.get_or_create(
    first_name='John',
    last_name='Lennon',
    defaults={'birthday': date(1940, 10, 9)},
)
```  
But in `ormar` we only have  `get_or_create(**kwargs)`.  
So we can't pass any defaults to create object with them.

MojixCoder:
I think I can work on this if you want.

collerek:
Sure, sounds like a good idea, but please use `_defaults` as a parameter name to allow word `defaults` as a column name in ormar/database

MojixCoder:
How about adding this feature to `update_or_create` as well?
I think all of the methods should have same API?

collerek:
Added in 0.11.0. Thanks! :)

PR #557, opened 2022-02-01
Update get_or_create method

MojixCoder:
[Issue #554](https://github.com/collerek/ormar/issues/554)

### DIFF
--- a/docs/queries/create.md
+++ b/docs/queries/create.md
@@ -63,6 +63,9 @@ Combination of create and get methods.
 Tries to get a row meeting the criteria and if `NoMatch` exception is raised it creates
 a new one with given kwargs and _defaults.
 
+When `_defaults` dictionary is provided the values set in `_defaults` will **always** be set, including overwriting explicitly provided values. 
+i.e. `get_or_create(_defaults: {"title": "I win"}, title="never used")` will always use "I win" as title whether you provide your own value in kwargs or not. 
+
 ```python
 class Album(ormar.Model):
     class Meta:
--- a/docs/queries/create.md
+++ b/docs/queries/create.md
@@ -3,7 +3,7 @@
 Following methods allow you to insert data into the database.
 
 * `create(**kwargs) -> Model`
-* `get_or_create(**kwargs) -> Model`
+* `get_or_create(_defaults: Optional[Dict[str, Any]] = None, **kwargs) -> Tuple[Model, bool]`
 * `update_or_create(**kwargs) -> Model`
 * `bulk_create(objects: List[Model]) -> None`
 
@@ -16,7 +16,7 @@ Following methods allow you to insert data into the database.
 
 * `QuerysetProxy`
       * `QuerysetProxy.create(**kwargs)` method
-      * `QuerysetProxy.get_or_create(**kwargs)` method
+      * `QuerysetProxy.get_or_create(_defaults: Optional[Dict[str, Any]] = None, **kwargs)` method
       * `QuerysetProxy.update_or_create(**kwargs)` method
 
 ## create
@@ -56,12 +56,12 @@ await malibu.save()
 
 ## get_or_create
 
-`get_or_create(**kwargs) -> Model`
+`get_or_create(_defaults: Optional[Dict[str, Any]] = None, **kwargs) -> Tuple[Model, bool]`
 
 Combination of create and get methods.
 
 Tries to get a row meeting the criteria and if `NoMatch` exception is raised it creates
-a new one with given kwargs.
+a new one with given kwargs and _defaults.
 
 ```python
 class Album(ormar.Model):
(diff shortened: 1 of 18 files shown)


ponytailer:
This change was incompatible for this method. Maybe need the big release version to merge it or a new method.

ormar/queryset/queryset.py
```diff
@@ -972,26 +972,36 @@ async def get(self, *args: Any, **kwargs: Any) -> "T":  # noqa: CCR001
...
         """
         try:
-            return await self.get(*args, **kwargs)
+            return await self.get(*args, **kwargs), False
         except NoMatch:
-            return await self.create(**kwargs)
+            if _defaults is None:
+                return await self.create(**kwargs), True
+            else:
+                return await self.create(**kwargs, **_defaults), True
```
collerek, line 1004: please use:
```python
_defaults = _defaults or {}
```
to avoid the if statement.

The order of arguments is wrong -> `_defaults` should be first, but still overwritten by `kwargs` value if explicitly provided.

Also, this will break if `kwargs` and `_default` have the same parameter. 
You are missing a test that checks this (that would fail). 
Even if you provide defaults you can as a user still pass a different value for this parameter, the way you did it would always overwrite it with the default one.

so it should be:
```python
return await self.create(**{**_defaults, **kwargs}), True
```
(update _defaults with kwargs and only then unpack).

ormar/relations/querysetproxy.py
```diff
@@ -483,23 +484,33 @@ async def update(self, each: bool = False, **kwargs: Any) -> int:
...
+        :rtype: Tuple("T", bool)
         """
         try:
-            return await self.get(*args, **kwargs)
-        except ormar.NoMatch:
-            return await self.create(**kwargs)
+            return await self.get(*args, **kwargs), False
+        except NoMatch:
+            if _defaults is None:
+                return await self.create(**kwargs), True
```
collerek, line 511: Same as above

tests/test_queries/test_queryset_level_methods.py
```diff
@@ -188,6 +189,28 @@ async def test_get_or_create():
             )
 
 
+@pytest.mark.asyncio
+async def test_get_or_create_with_defaults():
```
collerek, line 193: A missing path is same arguments in kwargs and in _defaults

collerek:
> This change was incompatible for this method. Maybe need the big release version to merge it or a new method.

Yes, this will wait for 0.11 release.

ormar/queryset/queryset.py
MojixCoder: should i raise an error if `_defaults` and `kwargs` have something in common?

tests/test_queries/test_queryset_level_methods.py
MojixCoder: Sorry I couldn't understand what do you mean

ormar/queryset/queryset.py
MojixCoder: I think this is fine for `_defaults` and `kwargs` to have some keys in common.  
It means get object with the given `kwargs`, if object was not found, then create it with given `_defaults` and `kwargs` but `_defaults` should overwrite `kwargs`.  
What do you think about this implementation?

collerek: OK, I forgot tit's get or create.
We can overwrite kwargs with defaults (and it will be in line with what django is doing).

Yes, it is fine to have the same key in kwargs and defaults (in your implementation currently they cannot), and we can agree defaults should overwrite the provided kwargs.

Also, I asked you to remove the if statement (the fewer conditionals the better), so you will always have defaults populated.

So full implementation will be like this:
```python
except NoMatch:
    _defaults = _defaults or {} # instantiate empty dict if defaults not provided)
    return await self.create(**{**kwargs, **_defaults}), True  # update kwargs dict with defaults and then unpack as kwargs to create
```

tests/test_queries/test_queryset_level_methods.py
collerek: right now you miss a test for `get_or_create(_defaults={"name":"test2"}, name="test")`.
Same argument in kwargs and defaults -> it will raise exception argument name provided multiple times to method create in you current implementation.

ormar/queryset/queryset.py
MojixCoder: Yeah I was waiting for your response.  
I'll make the changes and let you know.

MojixCoder: Changes have been made. @collerek

tests/test_queries/test_queryset_level_methods.py
MojixCoder: I've added a test for overwriting `kwargs` by `_defaults`.

MojixCoder:
hey,
Were the changes fine?

docs/queries/create.md
```diff
@@ -56,12 +56,12 @@ await malibu.save()
...
 ## get_or_create
 
-`get_or_create(**kwargs) -> Model`
+`get_or_create(_defaults: Optional[Dict[str, Any]] = None, **kwargs) -> Tuple[Model, bool]`
 
 Combination of create and get methods.
 
 Tries to get a row meeting the criteria and if `NoMatch` exception is raised it creates
-a new one with given kwargs.
+a new one with given kwargs and _defaults.
```
collerek, line 64: Can you please add a line explaining the preceedence of `_defaults`:

```
When `_defaults` dictionary is provided the values set in `_defaults` will **always** be set, including overwriting explicitly provided values. i.e. `get_or_create(_defaults: {"title": "I win"}, title="never used")` will always use "I win" as title whether you provide your own value in kwargs or not. 
```

PR #575, opened 2022-02-25
fix: IndexConstraint should pass kw-args to sqlalchemy

zevisert:
Title. We have to import sqlalchemy in order to pass other kwags through to the index


collerek:
Awesome, thanks! :)

PR #589, opened 2022-03-09
Fix collerek/ormar#588 Bug in queryset count() method

### LINKED ISSUE #588: Using "count()" on table with `select_related` gives the total of all rows instead of the count of the primary select
**Describe the bug**
This is mainly apparent in pagination, where you want the count of the primary objects. The count currently takes the count of all rows, which will include the joined rows. So if an object has many related objects, the count will be the number of related objects.

**To Reproduce**
I have added a test to check count in the "test_aggr" file on my forked branch, as well as some potential code to fix the issue.
I wasn't sure if you wanted to introduce a breaking change, or if the original functionality was necessary for something. I ended up changing the default, but adding a "distinct" keyword for those who depend on the legacy behavior.

**Expected behavior**
Count should return the count of the primary rows, not the total number of rows including left joins

**Versions (please complete the following information):**
 - Postgres 11.4
 - Python 3.8.11
 - ormar 0.10.22

erichaydel:
Fixes #588

### DIFF
--- a/README.md
+++ b/README.md
@@ -334,7 +334,7 @@ async def delete():
 
 async def joins():
     # Tho join two models use select_related
-    
+
     # Django style
     book = await Book.objects.select_related("author").get(title="The Hobbit")
     # Python style
@@ -348,7 +348,7 @@ async def joins():
     # By default you also get a second side of the relation
     # constructed as lowercase source model name +'s' (books in this case)
     # you can also provide custom name with parameter related_name
-    
+
     # Django style
     author = await Author.objects.select_related("books").all(name="J.R.R. Tolkien")
     # Python style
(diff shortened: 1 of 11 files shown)


collerek:
Thanks for catching that! :)
This is still breaking as the default behavior changes but soon we should release a 0.11 version anyway.

collerek:
Can you please update the docs too to include the new parameter (check the quick start in index, & readme, update queries/aggregations)?

erichaydel:
Done! Let me know how that looks

erichaydel:
It seems like it died immediately on Python 3.7, but I'm not sure why. Is there any good way to read the log output?

PR #688, opened 2022-06-03
Iterators QuerySet Method

SepehrBazyar:


SepehrBazyar:
Fixes #691

ormar/queryset/queryset.py
```diff
@@ -1055,6 +1056,37 @@ async def all(self, *args: Any, **kwargs: Any) -> List["T"]:  # noqa: A003
...
+        :type kwargs: Any
+        :return: asynchronous iterable generator of returned models
+        :rtype: AsyncGenerator[Model]
+        """
+
+        if kwargs or args:
+            return await self.filter(*args, **kwargs).iterator()
+
+        expr = self.build_select_expression()
+        rows = await self.database.fetch_all(expr)
```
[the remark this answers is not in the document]
SepehrBazyar: > You should be using `iterate` from `databases` and not `fetch_all`.
> 
> As of now, it's the same as you would call `iterate(Model.objects.all())` so there is one call to DB to fetch all records at once, and only then the iterator is created.

Hi @collerek 
Thank you for your attention.

Yes that's right
I understand what you mean and I will rewrite it.

SepehrBazyar:
Hi @collerek 
Changes were done.
Thank you for checking and approve and run.
Thanks.

ormar/queryset/queryset.py
```diff
@@ -1055,6 +1056,38 @@ async def all(self, *args: Any, **kwargs: Any) -> List["T"]:  # noqa: A003
...
+        """
+
+        if kwargs or args:
+            async for result in self.filter(*args, **kwargs).iterator():
+                yield result
+            return
+
+        expr = self.build_select_expression()
+        async for row in self.database.iterate(query=expr):
+            result_row = self._process_query_result_rows([row])
```
collerek, line 1085: Now that I think about this - it won't work for models with relations in this way.

Since in the case of a join one model is spread i.e. through three database rows (so 3 child models) if you iterate like this you would get three times the same parent with a different child each time. 

We had a similar problem with limit and offset queries where you should get one full parent model if you limit 1 and not only the first database row which is not the entire parent model.

SepehrBazyar: Hello again @collerek .
Thank you for reading and reviewing.

Yes, you are absolutely right;
But we still have this problem in limit and offset:
For example, when you want to limit the results with limit, if you do not True `limit_raw_sql`, we get an error,
And if we set its value to True, the restriction applies to the number of child models, not to the parent!

The same thing happens in the iterator,
which is repeated for the number of child models of the parent model;
But it works perfectly well on unrelated models.

Create an issue for the problem I mentioned?
What solution do you suggest to solve this problem?

SepehrBazyar: > Now that I think about this - it won't work for models with relations in this way.
> 
> Since in the case of a join one model is spread i.e. through three database rows (so 3 child models) if you iterate like this you would get three times the same parent with a different child each time.
> 
> We had a similar problem with limit and offset queries where you should get one full parent model if you limit 1 and not only the first database row which is not the entire parent model.

Hello again @collerek 
Thank you for reading and reviewing.

Yes, you are absolutely right;
But we still have this problem in limit and offset:
For example, when you want to limit the results with limit, if you do not True `limit_raw_sql`, we get an error,
And if we set its value to True, the restriction applies to the number of child models, not to the parent!

The same thing happens in the iterator,
which is repeated for the number of child models of the parent model;
But it works perfectly well on unrelated models.

Create an issue for the problem I mentioned?
What solution do you suggest to solve this problem?

collerek: What error do you get? It should work fine, so if there is some error it is probably a bug. Yes, please create an issue for it if you found something. Thanks 🙂

SepehrBazyar: > What error do you get? It should work fine, so if there is some error it is probably a bug. Yes, please create an issue for it if you found something. Thanks 🙂

I explained it in issue #704

SepehrBazyar: Hi again @collerek 
Thank you for fixing issue #704 

I reviewed the relevant codes again and finally came to the following conclusion.
Thank you for your help and suggestions.

We have two choices for the iterator:
1. Or get all the rows together with `fetch_all` and then write the model build as an async generator to be able to join.
2. The second approach is that if we want to use the `iterate` method of the `databases` package, because one line is received each time, we will not be able to use commands such as `select_related` and etc. when using this method.

collerek: Hi, sorry for taking a long time to review.

I belive we can make it work with select related (yield from db iterator until pk of the main model changes and only then yield from ormar).

That will require some additional work so if you need this urgently you can modify the QuerySet used and simply add your method to queryset and use it with the current version. Modifing the queryset is described here: https://collerek.github.io/ormar/models/#overwriting-the-default-queryset

So basically you can do:

```python
class IterableQuerySet(QuerySet):

    async def iterator(  # noqa: A003
        self,
        *args: Any,
        **kwargs: Any,
    ) -> AsyncGenerator["T", None]:
        """
        Return async iterable generator for all rows from a database for given model.
        Passing args and/or kwargs is a shortcut and equals to calling
        `filter(*args, **kwargs).iterator()`.
        If there are no rows meeting the criteria an empty async generator is returned.
        :param kwargs: fields names and proper value types
        :type kwargs: Any
        :return: asynchronous iterable generator of returned models
        :rtype: AsyncGenerator[Model]
        """

        if kwargs or args:
            async for result in self.filter(*args, **kwargs).iterator():
                yield result
            return

        expr = self.build_select_expression()
        async for row in self.database.iterate(query=expr):
            result_row = self._process_query_result_rows([row])
            if self._prefetch_related and result_row:
                result_row = await self._prefetch_related_models(result_row, [row])

            yield result_row[0]


class YourModel(ormar.Model):

    class Meta(ormar.ModelMeta):
        ...
        queryset_class = IterableQuerySet
    ...

```

collerek:
You are missing tests with select related (and oreferch related) for iterator now so the coverage is below 100%

ormar/queryset/queryset.py
SepehrBazyar: thank you for answering 
I did it by what you suggested and I am relieved to say It worked! I hadn't thought of this
So sorry to bother you and take so much of your time 

I tried so many times and all I got were errors! Therefore sorry that I committed  so many times 

Thanks for your help

collerek:
Ok, I simplified the checks as processing the rows is quite expensive and it's enough to check the primary key of the row alone.

Changed the name to `iterate` to be more in line with other libraries and the function of this method.

To do:
* The same method should be also on querysetproxy so you can iterate on related models from the parent model (that methods only apply filters and delegate to queryset). (and test for it)
* Write a test with a different primary key field type (i.e. string)

Great job and looking forward to your changes.

SepehrBazyar:
> Ok, I simplified the checks as processing the rows is quite expensive and it's enough to check the primary key of the row alone.
> 
> Changed the name to `iterate` to be more in line with other libraries and the function of this method.
> 
> To do:
> 
> * The same method should be also on querysetproxy so you can iterate on related models from the parent model (that methods only apply filters and delegate to queryset). (and test for it)
> * Write a test with a different primary key field type (i.e. string)
> 
> Great job and looking forward to your changes.

Yes, this is much more efficient, thank you.
The items you mentioned have been successfully added.
Also, with your permission, I updated the documents.

Now "CodeFactor" is failing, I don't know the reason;
I would be grateful if you could help me to fix it and tell me the reason,
thanks.

collerek:
It's complaining that the test_models became too complex as a file.

Can you extract all new tests into test_iterate.py file in the same module (test_model_definition)

ormar/queryset/queryset.py
```diff
@@ -1055,6 +1056,54 @@ async def all(self, *args: Any, **kwargs: Any) -> List["T"]:  # noqa: A003
...
+        *args: Any,
+        **kwargs: Any,
+    ) -> AsyncGenerator["T", None]:
+        """
+        Return async iterable generator for all rows from a database for given model.
+
+        Passing args and/or kwargs is a shortcut and equals to calling
+        `filter(*args, **kwargs).iterate()`.
+
+        If there are no rows meeting the criteria an empty async generator is returned.
```
collerek, line 1070: This does not work.

You check with an empty result (list index exception):

```python
@pytest.mark.asyncio
async def test_empty_result():
    async with database:
        async with database.transaction(force_rollback=True):
            async for user in User.objects.iterate():
                pass
```

ormar/queryset/queryset.py
```diff
@@ -1055,6 +1056,54 @@ async def all(self, *args: Any, **kwargs: Any) -> List["T"]:  # noqa: A003
...
+            if last_primary_key == current_primary_key or last_primary_key is None:
+                last_primary_key = current_primary_key
+                rows.append(row)
+                continue
+
+            yield self._process_query_result_rows(rows)[0]
+            last_primary_key = current_primary_key
+            rows = [row]
+
+        yield self._process_query_result_rows(rows)[0]
```
collerek, line 1105: This will raise exception on empty result

docs/queries/read.md
```diff
@@ -173,6 +173,45 @@ tracks = await Track.objects.all()
...
+        database = database
+
+    id: int = ormar.Integer(primary_key=True)
+    name: str = ormar.String(max_length=100)
+```
+
+```python
+await Album.objects.create(name='The Cat')
+await Album.objects.create(name='The Dog')
+# will Asynchronous Iterable Generator for All Rows Album Model:
```
collerek, line 200: Please change the description in a comment to:
`will asynchronously iterate all Album models yielding one main model at a time from the generator`

collerek (commented):
All good except minor wording and empty result set.

collerek:
Awesome, thanks for your contribution! :)

SepehrBazyar:
Hi @collerek 
In order to add this feature, a new version of ormar needs to be released so that these commands and methods can be used.
Thanks

Issue #699, opened 2022-06-11
Enum leads to ValueError when loading OpenAPI-Spec with FastAPI


mawoka-myblock:
**Describe the bug**
If a model includes an Enum, everything works as expected untill I try to open the OpenAPI-Specs generated by FastAPI.

**To Reproduce**
Create a project with ormar and FastAPI (I also use Alembic). Create the following modell:

```python
class UserAuthTypes(Enum):
    LOCAL = "LOCAL"
    GOOGLE = "GOOGLE"


class User(ormar.Model):
    id: uuid.UUID = ormar.UUID(primary_key=True, default=uuid.uuid4())
    email: str = ormar.String(unique=True, max_length=100)
    username: str = ormar.String(unique=True, max_length=100)
    password: str = ormar.String(max_length=100)
    verified: bool = ormar.Boolean(default=False)
    verify_key: str = ormar.String(unique=True, max_length=100, nullable=True)
    created_at: datetime = ormar.DateTime(default=datetime.now())
    auth_type: UserAuthTypes = ormar.Enum(enum_class=UserAuthTypes, default=UserAuthTypes.LOCAL)
    google_uid: Optional[str] = ormar.String(unique=True, max_length=25, nullable=True)

    class Meta:
        tablename = "users"
        metadata = metadata
        database = database

    class Config:
        use_enum_values = True
```

Then, execute the following SQL to tell Postgres what this "userauthtypes" is:
```sql
CREATE TYPE userauthtypes AS ENUM ('LOCAL', 'GOOGLE');
```
Open the OpenAPI-Spec from FastAPI and see that it crashes. You may need to have the User-model as an input/response-class.

**Expected behavior**
It should work.

**Screenshots**
<details>
<summary>The following error occurs </summary>

```text
Traceback (most recent call last):
  File "/home/mawoka/.local/share/virtualenvs/ClassQuiz-BX_cFwCN/lib/python3.10/site-packages/starlette/middleware/base.py", line 43, in call_next
    message = await recv_stream.receive()
  File "/home/mawoka/.local/share/virtualenvs/ClassQuiz-BX_cFwCN/lib/python3.10/site-packages/anyio/streams/memory.py", line 94, in receive
    return self.receive_nowait()
  File "/home/mawoka/.local/share/virtualenvs/ClassQuiz-BX_cFwCN/lib/python3.10/site-packages/anyio/streams/memory.py", line 87, in receive_nowait
    raise EndOfStream
anyio.EndOfStream

[...]

PR #714, opened 2022-06-25
Added failing Enum-test

mawoka-myblock:
This test proofs https://github.com/collerek/ormar/issues/699


collerek:
Thanks for the investigation - included more verbose test in the code, thanks! 😁

PR #730, opened 2022-07-05
CheckColumns Constraint

### LINKED ISSUE #729: CheckColumns Constraint
```python
import datetime
import databases
import sqlalchemy

import ormar

database = databases.Database("sqlite:///db.sqlite")
metadata = sqlalchemy.MetaData()


class Course(ormar.Model):
    class Meta:
        database = database
        metadata = metadata
        # define your constraints in Meta class of the model
        # it's a list that can contain multiple constraints
        # hera a combination of name and column will have a level check in the db
        constraints = [
            ormar.CheckColumns("start_time < end_time", name="date_check"),
        ]

    id: int = ormar.Integer(primary_key=True)
    name: str = ormar.String(max_length=100)
    start_date: datetime.date = ormar.Date()
    end_date: datetime.date = ormar.Date()
```

SepehrBazyar:
fixes #729

### DIFF
--- a/ormar/models/helpers/sqlalchemy.py
+++ b/ormar/models/helpers/sqlalchemy.py
@@ -318,7 +318,7 @@ def set_constraint_names(meta: "ModelMeta") -> None:
                 f'{"_".join([col for col in constraint._pending_colargs])}'
             )
         elif isinstance(constraint, sqlalchemy.CheckConstraint) and not constraint.name:
-            sql_condition: str = constraint.sqltext.replace(" ", "_")
+            sql_condition: str = str(constraint.sqltext).replace(" ", "_")
             constraint.name = f"check_{meta.tablename}_{sql_condition}"
 
 
--- a/ormar/models/helpers/sqlalchemy.py
+++ b/ormar/models/helpers/sqlalchemy.py
@@ -298,7 +298,7 @@ def populate_meta_sqlalchemy_table_if_required(meta: "ModelMeta") -> None:
 
 def set_constraint_names(meta: "ModelMeta") -> None:
     """
-    Populates the names on IndexColumn and UniqueColumns constraints.
+    Populates the names on IndexColumns and UniqueColumns and CheckColumns constraints.
 
     :param meta: Meta class of the Model without sqlalchemy table constructed
     :type meta: Model class Meta
@@ -316,6 +316,11 @@ def set_constraint_names(meta: "ModelMeta") -> None:
             constraint.name = (
                 f"ix_{meta.tablename}_"
                 f'{"_".join([col for col in constraint._pending_colargs])}'
+            )
+        elif isinstance(constraint, sqlalchemy.CheckConstraint) and not constraint.name:
+            constraint.name = (
+                f"cc_{meta.tablename}_"
+                f'{"_".join([str(col) for col in constraint._pending_colargs])}'
             )
 
 
--- a/ormar/models/helpers/sqlalchemy.py
+++ b/ormar/models/helpers/sqlalchemy.py
@@ -318,10 +318,7 @@ def set_constraint_names(meta: "ModelMeta") -> None:
                 f'{"_".join([col for col in constraint._pending_colargs])}'
             )
         elif isinstance(constraint, sqlalchemy.CheckConstraint) and not constraint.name:
-            constraint.name = (
-                f"cc_{meta.tablename}_"
-                f'{"_".join([str(col) for col in constraint._pending_colargs])}'
-            )
+            constraint.name = f"check_{meta.tablename}_{constraint.sqltext}"
 
 
 def update_column_definition(
--- a/ormar/models/helpers/sqlalchemy.py
+++ b/ormar/models/helpers/sqlalchemy.py
@@ -318,7 +318,8 @@ def set_constraint_names(meta: "ModelMeta") -> None:
                 f'{"_".join([col for col in constraint._pending_colargs])}'
             )
         elif isinstance(constraint, sqlalchemy.CheckConstraint) and not constraint.name:
-            constraint.name = f"check_{meta.tablename}_{constraint.sqltext}"
+            sql_condition: str = constraint.sqltext.replace(" ", "_")
+            constraint.name = f"check_{meta.tablename}_{sql_condition}"
 
 
 def update_column_definition(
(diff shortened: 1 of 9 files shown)


SepehrBazyar:
> Awesome, thanks! :)
> 
> There are a few missing steps here.
> 
> 1. All constraints in ormar have names defined by default (if you do not provide your own explicitly). The reason is related to alembic migrations which will otherwise constantly re-declare the indexes instead of keeping the old ones. The name assignment is here: https://github.com/collerek/ormar/blob/0f2f4badbc4c2acca932f393d421e3e26313b304/ormar/models/helpers/sqlalchemy.py#L299-L320
> 2. The constraints are copied when there are child models declared. BTW. I see it was missed for IndexColumns already and needs to be adjusted for both `IndexColumns` and `CheckColumns` constraints.
>    So you need to add tests for `CheckColumns` in the parent model and see if they are populated in child models. Since you already do this I would be grateful if you could also include `IndexColumns`. Attributes from meta are copied here: https://github.com/collerek/ormar/blob/0f2f4badbc4c2acca932f393d421e3e26313b304/ormar/models/metaclass.py#L198-L232
>    
>    And as you can see it will overwrite all constraints with `UniqueColumns` cause it supports neither `IndexColumns` nor `CheckColumns`. Inheritance tests are in test_inheritance_concrete.py file.

Hi @collerek 
The changes have been applied. Thank you for checking and giving your opinion about their correctness.
Now the tests run correctly on mysql and sqlite, but it fails in postgres, which I don't know why.

ormar/models/helpers/sqlalchemy.py
```diff
@@ -317,6 +317,8 @@ def set_constraint_names(meta: "ModelMeta") -> None:
                 f"ix_{meta.tablename}_"
                 f'{"_".join([col for col in constraint._pending_colargs])}'
             )
+        elif isinstance(constraint, sqlalchemy.CheckConstraint) and not constraint.name:
+            constraint.name = f"check_{meta.tablename}_{constraint.sqltext}"
```
collerek, line 321: In some backends names cannot contain special characters and spaces so it cannot be like this, you have to process is to keep only [a-zA-z_]

ormar/models/metaclass.py
```diff
@@ -222,10 +222,6 @@ def update_attrs_from_base_meta(  # noqa: CCR001
                     model_fields=model_fields,
                     parent_value=parent_value,
                 )
-                parent_value = [
-                    ormar.fields.constraints.UniqueColumns(*x._pending_colargs)
-                    for x in parent_value
-                ]
```
collerek, line 228: You cannot remove this piece, this is why some of the postgress fails.

Basically, you have to copy the constraint unpacking it's values (here and for index it is `_pending_colargs`, not sure for CheckColumns. If you take the parent value as is (as it is now) the constraint is already bound to the previously created already existing table and will fail.

ormar/models/metaclass.py
```diff
@@ -195,6 +196,29 @@ def verify_constraint_names(
...
+    """
+
+    if isinstance(value, sqlalchemy.UniqueConstraint):
+        return UniqueColumns(*value._pending_colargs)  # pragma: no cover
+    elif isinstance(value, sqlalchemy.Index):
+        return IndexColumns(*value._pending_colargs)  # pragma: no cover
+    elif isinstance(value, sqlalchemy.CheckConstraint):
+        return CheckColumns(value.sqltext)  # pragma: no cover
+
+    raise ValueError(f"{value} Must be a ColumnCollectionConstraint.")
```
collerek, line 219: This can be re-written as:

```python
def get_constraint_copy(
    constraint: ColumnCollectionMixin # I think index is not a constraint in sqlalchemy and Mixin is the least common part?
) -> Union[UniqueColumns, IndexColumns, CheckColumns]::
    constraints = {
        sqlalchemy.UniqueConstraint: lambda x: UniqueColumns(*x._pending_colargs),
        sqlalchemy.Index: lambda x: IndexColumns(*x._pending_colargs),
        sqlalchemy.CheckConstraint: lambda x: CheckColumns(x.sqltext),
    }
    checks = (key if isinstance(constraint, key) else None for key in constraints)
    target_class = next((target for target in checks if target is not None), None)
    constructor = constraints.get(target_class)
    if not constructor:
        raise ValueError(f"{constraint} must be a ColumnCollectionMixin!")
    return constructor(constraint)
```

It seems more complicated but it has number of advantages.
1. You can see straight away what are all your options in constraints
2. You can easily expand by adding new entries to constraints dictionaries without changing anything else
3. You avoid complicated if/else statements
4. You avoid pragma no cover comments

tests/test_meta_constraints/test_check_constraints.py
```diff
@@ -0,0 +1,57 @@
...
+
+@pytest.mark.asyncio
+async def test_check_columns():
+    async with database:
+        async with database.transaction(force_rollback=True):
+            await Product.objects.create(
+                name="Mars", company="Nestle", inventory=100, buffer=10
+            )
+
+            if Product.Meta.database._backend._dialect.name != "mysql":
```
collerek, line 48: You could pragma no cover the whole if statement instead of both of children statements separately

ormar/models/metaclass.py
SepehrBazyar: Yes, that's right, it was perfect
Thank you for your guidance

collerek:
Well done!
Awesome job, let's now tackle your last PR and we will make a new release with those changes :)
Thanks!

ormar/models/metaclass.py
```diff
@@ -195,6 +197,33 @@ def verify_constraint_names(
...
+    constraints = {
+        sqlalchemy.UniqueConstraint: lambda x: UniqueColumns(*x._pending_colargs),
+        sqlalchemy.Index: lambda x: IndexColumns(*x._pending_colargs),
+        sqlalchemy.CheckConstraint: lambda x: CheckColumns(x.sqltext),
+    }
+    checks = (key if isinstance(constraint, key) else None for key in constraints)
+    target_class = next((target for target in checks if target is not None), None)
+    constructor: Optional[Callable] = constraints.get(target_class)
+    if not constructor:
+        raise ValueError(f"{constraint} must be a ColumnCollectionMixin!")
```
zevisert, line 222: If this is the case why not use `issubclass(target_class, ColumnCollectionMixin)`?

collerek: As you see we need a different attribute of the "parent" (either `_pending_colargs` or `sqltext`). 

You have to unpack the constraints attributes otherwise it will not work with inheritance as the constraints would have been bound to the parent model and fail. All of those (unique, index and check) constraints are subclasses od `ColumnCollectionMixin` so it's not enough as a check.

PR #735, opened 2022-07-07
Referential Actions Enum Class

### LINKED ISSUE #724: ForeignKeyViolationError in CASCADE clear() method
**Describe the bug:**
When I try to CASCADE delete relational models.

**To Reproduce:**
```python
# Models
import ormar
from uuid import UUID, uuid4

class A(ormar.Model):
    id: UUID = ormar.UUID(uuid_format="string", primary_key=True, default=uuid4)
    name: str = ormar.String(max_length=64, nullalbe=False)

class B(ormar.Model):
    id: UUID = ormar.UUID(uuid_format="string", primary_key=True, default=uuid4)
    name: str = ormar.String(max_length=64, nullalbe=False)
    a: A = ormar.ForeignKey(to=A)

class C(ormar.Model):
    id: UUID = ormar.UUID(uuid_format="string", primary_key=True, default=uuid4)
    name: str = ormar.String(max_length=64, nullalbe=False)
    b: B = ormar.ForeignKey(to=B)

# QuerySet
# (a is an object from A model class.)
>>> await a.bs.clear(keep_reversed=False)
~/Desktop/MyProject/venv/lib/python3.10/site-packages/asyncpg/protocol/protocol.pyx in bind_execute()

ForeignKeyViolationError: update or delete on table "bs" violates foreign key constraint "fk_cs_bs_id_b" on table "cs"
DETAIL:  Key (id)=(592c8bd6-6796-48d0-9158-fe1fcd0cfd57) is still referenced from table "cs".
```

**Versions:**
 - Database backend: PostgreSQL
 - Python version: 3.10.4
 - `ormar` version: 0.11.2
 - `pydantic` version: 1.9.0

SepehrBazyar:
fixes #724


SepehrBazyar:
Hi @collerek 
Since the `default` parameter is not allowed in the definition of `ForeignKey`, will the `SET DEFAULT` action find meaning?
Thank you for checking and telling me what you think should be done for this case.

SepehrBazyar:
Hello, dear @collerek 
Thank you for reviewing and giving your opinion

tests/test_relations/test_database_fk_creation.py
```diff
@@ -53,3 +81,25 @@ def test_simple_cascade():
...
+
+@pytest.mark.asyncio
+async def test_cascade_clear():
+    async with database:
+        async with database.transaction(force_rollback=True):
+            a = await A.objects.create(name="a")
+            b = await B.objects.create(name="b", a=a)
+            c = await C.objects.create(name="c", b=b)
+
+            await a.bs.clear(keep_reversed=False)
```
collerek, line 105: Add check that they are really gone please

ormar/actions.py
```diff
@@ -0,0 +1,26 @@
...
+    those references have to be considered.
+
+    SQL specifies 5 different referential actions
+    that shall take place in such occurrences.
+    """
+
+    CASCADE: str = "CASCADE"
+    RESTRICT: str = "RESTRICT"
+    SET_NULL: str = "SET NULL"
+    SET_DEFAULT: str = "SET DEFAULT"
```
collerek, line 25: I would leave that despite you are not really allowed to set default on fk in ormar, you can still do it in database directly

collerek (changes requested):

ormar/actions.py
```diff
@@ -0,0 +1,26 @@
+"""
+Gathers all referential actions by ormar.
+"""
+
+
+from enum import Enum
+
+
+class Action(Enum):
```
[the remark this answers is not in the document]
SepehrBazyar: Yes, you are right and I did
I am asking just to be sure, do you mean `Referencial` the same as `Referential` or are they different?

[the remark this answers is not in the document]
collerek: Yeah it's a typo, sorry referential that is

ormar/actions.py
SepehrBazyar: My question is why we are not allowed to set `default` in ormar?
If we add this possibility, what problem might occur?

collerek: Because the default values in pydantic are assigned in init, but init cannot be async.
You cannot save anything to the database in init (bad practice anyway). 
That means that the default value set on the foreign key would have to already exist in db prior to model declaration, it won't be created with the parent model. 

That would be super confusing for users that would want to add a whole related model as default in the parent model declaration (wouldn't work). And would require creating and **saving** the default model in db before the model that uses the it as default value.

SepehrBazyar: Excellent, I completely understand
So, I will add a note in the document that to use `SET_DEFAULT`, the `server_default` must be specified.

collerek (approved):

collerek:
Awesome, thanks! :)

PR #769, opened 2022-07-31
Nulls Last and Nulls First Parameters Sorting

SepehrBazyar:


SepehrBazyar:
Hi @collerek 
Thank you, please, if you do the review.

collerek:
Since both options are mutually exclusive (you cannot have nulls both last and first at the same time) that should be a single option, something like: `nulls_ordering` that accepts an Enum like `Nulls.LAST` or `Nulls.FIRST` and `Nulls.UNSET` as default (or `None`). 

That would also greatly simplify the conditions in the code as you have explicit value to handle.

ormar/queryset/actions/order_action.py
```diff
@@ -95,6 +111,40 @@ def _split_value_into_parts(self, order_str: str) -> None:
...
+        :type result: str
+        :return: result of the final query by field name and direction and nulls value
+        :rtype: str
+        """
+
+        if not self.is_mysql_bool:
+            return result + f" nulls {self.nulls}"  # pragma: no cover
+
+        condition: str = "not" if self.nulls == "first" else ""  # pragma: no cover
+        return f"{field_name} is {condition} null, {result}"  # pragma: no cover
```
collerek, line 130: If you split this condition into two the result now lacks prefix and will not work with nested models.

You would have in example `{prefix}{table_name}.{field_name} is null, {field_name} {self.direction}`.

The second field_name does not have the prefix.

tests/test_queries/test_order_by.py
```diff
@@ -163,6 +163,50 @@ async def test_sort_order_on_main_model():
         assert songs[2].name == "Song 2"
         assert songs[3].name == "Song 3"
 
+        await Song.objects.create(name="Song 5")
```
collerek, line 166: There are missing tests for sorting by related model field (so sorting songs by album.title or whatever)

ormar/queryset/actions/order_action.py
```diff
@@ -95,6 +111,40 @@ def _split_value_into_parts(self, order_str: str) -> None:
         self.field_name = parts[-1]
         self.related_parts = parts[:-1]
 
+    def _handle_field_nulls_mysql(self, field_name: str, result: str) -> str:
```
collerek, line 114: name is misleading as it handles all dialects and not only mysql

ormar/queryset/field_accessor.py
```diff
@@ -268,22 +268,40 @@ def isnull(self, other: Any) -> FilterGroup:
         """
         return self._select_operator(op="isnull", other=other)
 
-    def asc(self) -> OrderAction:
+    def asc(self, nulls_ordering: Optional[NullsOrdering] = None) -> OrderAction:
```
collerek, line 271: You should also be able to pass this param in `order_by` in queryset and querysetproxy for related models.

collerek (changes requested):

ormar/queryset/field_accessor.py
SepehrBazyar: > You should also be able to pass this param in `order_by` in queryset and querysetproxy for related models.

I do not understand this
How to determine this amount in `order_by()` and without `asc()` or `desc()`?

ormar/queryset/actions/order_action.py
SepehrBazyar: > If you split this condition into two the result now lacks prefix and will not work with nested models.
> 
> You would have in example `{prefix}{table_name}.{field_name} is null, {field_name} {self.direction}`.
> 
> The second field_name does not have the prefix.

Of course, I must mention that without this case, the test related to nested models will also pass correctly. Can you give an example of a possible error that may occur? Thanks

ormar/queryset/field_accessor.py
collerek: In order by you can pass OrderActions or strings too. If you pass `sort_order` it's ascending, if you pass `-sort_order` it's descending. For nested models double underscore is required so like `main_model_relation_field__child_model_field`. It also supports putting a `-` before nested related model fields and then it's descending. The logic is starting in order_by where OrderActions are created from strings.

ormar/queryset/actions/order_action.py
collerek: When you have a parent model with two relations to the child model. So like items with `created_by` and `modified_by` that lead to the same `User` model. Each of those relations in joins has to have different prefixes as otherwise, you would have ambiguous column names.

ormar/queryset/actions/order_action.py
```diff
@@ -30,10 +35,16 @@ def __init__(
         if self.source_model == self.target_model and "__" not in self.related_str:
             self.is_source_model_order = True
 
+        self.nulls = nulls_ordering if nulls_ordering is not None else None
```
collerek, line 38: that line should be simply `self.nulls = nulls_ordering` cause `is not None else None` part means that if `null_ordering` is None it's gonna be None anyway. You should also probably use `self.nulls_ordering` variable name to store the value.

SepehrBazyar: Hi @collerek 
I apologize for the delay;
I made the changes you said.

collerek (approved):

Issue #843, opened 2022-09-23
bulk_create does not check empty list before inserting


Abdeldjalil-H:
The method `QuerySet.bulk_create` does not check if the `objects` list is empty or not before inserting. If the list is empty, the insert query is still executed. This will raise an error if you're inserting to a table that has a column without default values. Otherwise a row will be added to the database.
I think the method should check whether the `objects` list empty or not before inserting.

PR #853, opened 2022-09-27
support(bulk-create): add raise_error flag when bulk-create empty models

ponytailer:
- add raise_error flag #843

### DIFF
--- a/docs/relations/queryset-proxy.md
+++ b/docs/relations/queryset-proxy.md
@@ -127,7 +127,7 @@ provided Through model.
 
 Given sample like this:
 
-```Python hl_lines="14-20, 29"
+```Python hl_lines="14-20 29"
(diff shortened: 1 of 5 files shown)


Abdeldjalil-H (changes requested):
I think it is better to return when `raise_error=False` and the objects list is empty.

collerek:
@ponytailer We can do it inline with `bulk_update` and simply raise and exception on empty list (and release as 0.12.0 since it's breaking). Thanks!

ponytailer:
> ### **[collerek](/collerek) ** commented [2分钟前](#issuecomment-1260714795)

Yes, if you will have the big version recently,  raise exception simply when it's empty.

collerek (approved):

collerek:
Thanks!

PR #866, opened 2022-10-04
exclude null fields in _construct_relations

Abdeldjalil-H:
In the method `NewBaseModel._construct_relations` if some relations are null, `expand_relationship` will return `None`. This `None` values will cause the following `AttributeError` when they are passed to `model._orm.add` as `parent`

```
ormar/relations/relation_manager.py", in add
    parent_relation = parent._orm._get(child_name)
AttributeError: 'NoneType' object has no attribute '_orm'
```


Abdeldjalil-H:
Can you please delete this PR, so I can open a clean one because it contains unnecessary commits.

collerek:
In github you cannot delete the pr, I can close it for you so you can open a new one.

PR #870, opened 2022-10-07
excludee null fields in _construct_relations

Abdeldjalil-H:
In the method `NewBaseModel._construct_relations` if some relations are null, `expand_relationship` will return `None`. This `None` values will cause the following `AttributeError` when they are passed to `model._orm.add` as `parent`

```
ormar/relations/relation_manager.py", in add
    parent_relation = parent._orm._get(child_name)
AttributeError: 'NoneType' object has no attribute '_orm'
```


collerek:
Awesome, thanks!
Can you please add one test with `None` as one of the relations so it will be checked from now on?

collerek (changes requested):
There is a missing test that this change should fix.

PR #948, opened 2022-11-29
Add benchmarking test suite and greatly improve performance in a few cases

erichaydel:
First off, this is a really great library--my team is using it and loving its functionality and simplicity. Since we're using in production, we're finding a few issues when dealing with larger sets of data. They are:
1. When doing `all` with `select_related`, the models load with N^2 (or worse) time where N is the number of related models.
2. When creating lots of models with a shared related model, it slows down as N^2 with the number of models.

Here, I've submitted fixes for these two issues, as well as a complete benchmarking test suite. I tried to test most of the major queryset paths from the documentation. I also threw in a few cases that replicated some of the issues we were seeing.

Here's some of my original benchmarking--in the case of the `all` slowdown, I was seeing even like N^2.5 or N^3 performance. The Y axis here is seconds, and the X axis is number of related models.
graph

Blue the baseline. Orange is the speed with bb0bf8cef7e1ed888b3493e4eec6c0cb0550e55c applied. Gray is 8d8e0a9d83c6185b4d46db4757dc482e928f701a applied as well. Yellow is the the final commit 886bc270feee6c652e73ed2d533f9a37796ebb6d that gets it to O(N) time. If you want, I can squash these--I just wanted it to be clear the value of each change. I couldn't replicate these findings easily with a pet test case since there were a lot of related models and things, but I did replicate something like N^2.2, which should be good enough to address the problem.

A little about the benchmarks: There's something strange about pytest-asyncio that takes forever to run a lot of coroutines. For example, I have a `benchmark.py` script attached here. For 1000 rows, it takes about 0.2 seconds to save each individually. If I do the same as part of a pytest test, it'll take multiple minutes. Not sure why. Therefore, they're probably not great for figuring out absolute performance or speedups between functions--but they're definitely helpful in seeing if changes sped up or slowed down code between commits.

The parameterized N is in brackets. You can see that I'm doubling parameters between every test, so a doubling of mean time would indicate linear N time.

The two benchmarks that are most important are, in the benchmarks folder:
`test_initializing_models_with_related_models`
benchmark_pre_optimization_initialization_with_related_models

and `test_get_all_with_related_models`
benchmark_pre_optimization_get_all_with_related_models

You can see the `test_initializing_models_with_related_models` is slightly better than O(N^2). `test_get_all_with_related_models` is slightly worse than O(N^2), and takes multiple seconds for hundreds of relations.

Here are the two values after these changes are applied:
benchmark_post_optimization_initialization_with_related_models

benchmark_post_optimization_get_all_with_related_models

And here are the direct speedups for these two
speedup_initialize_with_related_models
speedup_get_all_with_related_models

We're getting a pretty large speed up, and we're out of the O(N^2) or O(N^2.5) range into the O(N) range.

So there's also one last change I included: for our team, it's also very important not to block the event loop. In ormar, there are a lot of places where pulling a lot of rows from the DB will block the event loop since it does a lot of list comprehension. I broke those up a small amount and added `await asyncio.sleep(0)`, which shouldn't slow us down since it's basically a noop that allows the event loop to take over if it has pressing tasks. I measured the blocking time via the `jitter` checking in the `benchmark.py` script I've included here--basically, a separate thread tries to wake up every 100ms. The lag in that determines the blocking time of the algorithms. Feel free to check it out. Overall, it's a pretty noninvasive change and it does a great job of breaking up the blocking tasks.

I've included all my benchmark files (they have ANSI colors, so you may have to pipe it or read it with an IDE that can process that), as well as a `benchmark.py` file I've been using to measure jitter. [](url)

[comparison.log](https://github.com/collerek/ormar/files/10116316/comparison.log)
[post_optimization.log](https://github.com/collerek/ormar/files/10116315/post_optimization.log)
[pre_optimization.log](https://github.com/collerek/ormar/files/10116312/pre_optimization.log)
[benchmark.py.txt](https://github.com/collerek/ormar/files/10109570/benchmark.py.txt)

Anyway, let me know what you think!

### DIFF
--- a/ormar/models/mixins/merge_mixin.py
+++ b/ormar/models/mixins/merge_mixin.py
@@ -36,11 +36,24 @@ class MergeModelMixin:
         for model in result_rows:
             grouped_instances.setdefault(model.pk, []).append(model)
 
+        def recursive_add(g):
+            if len(g) <= 1:
+                return g
+
+            added_values = []
+            iterable_group = iter(g)
+            for model in iterable_group:
+                next_model = next(iterable_group, None)
+                if next_model is not None:
+                    combined = cls.merge_two_instances(next_model, model)
+                else:
+                    combined = model
+                added_values.append(combined)
+
+            return recursive_add(added_values)
+
         for group in grouped_instances.values():
-            model = group.pop(0)
-            if group:
-                for next_model in group:
-                    model = cls.merge_two_instances(next_model, model)
+            model = recursive_add(group)[0]
             merged_rows.append(model)
 
         return merged_rows
--- a/ormar/models/mixins/merge_mixin.py
+++ b/ormar/models/mixins/merge_mixin.py
@@ -17,6 +17,23 @@ class MergeModelMixin:
     """
 
     @classmethod
+    def _recursive_add(cls, g: List["Model"]) -> List["Model"]:
+        if len(g) <= 1:
+            return g
+
+        added_values = []
+        iterable_group = iter(g)
+        for model in iterable_group:
+            next_model = next(iterable_group, None)
+            if next_model is not None:
+                combined = cls.merge_two_instances(next_model, model)
+            else:
+                combined = model
+            added_values.append(combined)
+
+        return cls._recursive_add(added_values)
+
+    @classmethod
     def merge_instances_list(cls, result_rows: List["Model"]) -> List["Model"]:
         """
         Merges a list of models into list of unique models.
@@ -36,24 +53,8 @@ class MergeModelMixin:
         for model in result_rows:
             grouped_instances.setdefault(model.pk, []).append(model)
 
-        def recursive_add(g):
-            if len(g) <= 1:
-                return g
-
-            added_values = []
-            iterable_group = iter(g)
-            for model in iterable_group:
-                next_model = next(iterable_group, None)
-                if next_model is not None:
-                    combined = cls.merge_two_instances(next_model, model)
-                else:
-                    combined = model
-                added_values.append(combined)
-
-            return recursive_add(added_values)
-
         for group in grouped_instances.values():
-            model = recursive_add(group)[0]
+            model = cls._recursive_add(group)[0]
             merged_rows.append(model)
 
         return merged_rows
--- a/ormar/models/mixins/merge_mixin.py
+++ b/ormar/models/mixins/merge_mixin.py
@@ -17,12 +17,20 @@ class MergeModelMixin:
     """
 
     @classmethod
-    def _recursive_add(cls, g: List["Model"]) -> List["Model"]:
-        if len(g) <= 1:
-            return g
+    def _recursive_add(cls, model_group: List["Model"]) -> List["Model"]:
+        """
+        Instead of accumulating the model additions one by one, this recursively adds
+        the models. E.G.
+        [1, 2, 3, 4].accumulate_add() would give [3, 3, 4], then [6, 4], then [10]
+        where this method looks like
+        [1, 2, 3, 4].recursive_add() gives [[3], [7]], [10]
+        It's the same number of adds, but it gives better O(N) performance on sublists
+        """
+        if len(model_group) <= 1:
+            return model_group
 
         added_values = []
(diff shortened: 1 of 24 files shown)


ormar/models/newbasemodel.py
```diff
@@ -171,12 +172,29 @@ def __setattr__(self, name: str, value: Any) -> None:  # noqa CCR001
...
         else:
             # let pydantic handle errors for unknown fields
             super().__setattr__(name, value)
 
+        new_hash = hash(self)
+        if prev_hash == new_hash:
+            return
+
+        # Update all relation proxy caches with different hash
+        for relation in self._orm._relations.values():
```
erichaydel, line 188: For review: There might be a better way to do this. Basically, I just wanted to make sure to update the cache when the current model changes--this code block basically makes sure it either updates its own cache, or updates the cache of its related model when it's in the related model list.

ormar/relations/relation.py
```diff
@@ -130,13 +130,20 @@ def _find_existing(
             raise ValueError("Cannot find existing models in parent relation type")
         if self._to_remove:
             self._clean_related()
-        for ind, relation_child in enumerate(self.related_models[:]):
+
+        if child not in self.related_models:
+            return None
+        else:
+            # We need to clear the weakrefs that don't point to anything anymore
+            # There's an assumption here that if some of the related models went out of scope,
```
erichaydel, line 138: I put a comment, but I just wanted to highlight it: before, every time it checked for related models, it would check every item in the list for weakrefs. That leads to N^2 performance since for every related model you add, you have to run through the whole list. To get around this, I changed it to basically only check the first item and see if it's a weakref. If it is, check the whole list--otherwise, continue on.

The assumption is that there won't be weakrefs that go out of scope randomly in the middle of the list. As far as I can tell, that's a good assumption, but I wanted to ask about it.

ormar/models/newbasemodel.py
```diff
@@ -213,6 +223,26 @@ def __setstate__(self, state: Dict[Any, Any]) -> None:
         for name, value in relations.items():
             setattr(self, name, value)
 
+    def _update_relation_cache(self, prev_hash: int, new_hash: int) -> None:
```
erichaydel, line 226: For review: There might be a better way to do this. Basically, I just wanted to make sure to update the cache when the current model changes--this code block basically makes sure it either updates its own cache, or updates the cache of its related model when it's in the related model list. 

It's slightly inefficient since it doesn't actually check that the foreign relation list is of the same type as the current model--but I wasn't sure how to do that. The list check is O(1) time so I didn't think it would be too big of an issue.

collerek:
That looks impressive! 🎉 
Will try to review over the weekend, great job!

ormar/relations/relation_proxy.py
```diff
@@ -55,6 +79,55 @@ def related_field_name(self) -> str:
     def __getitem__(self, item: Any) -> "T":  # type: ignore
         return super().__getitem__(item)
 
+    def append(self, item: "T") -> None:
+        idx = len(self)
+        self._relation_cache[item.__hash__()] = idx
+        super().append(item)
+
+    def update_cache(self, prev_hash: int, new_hash: int) -> None:
```
collerek, line 87: Please provide the missing docstrings.

collerek (changes requested):

collerek:
@erichaydel Great job!🚀 
Some minor adjustments and we are good to merge and release the new version!

ormar/models/mixins/merge_mixin.py
```diff
@@ -16,6 +16,23 @@ class MergeModelMixin:
     in the end all parent (main) models should be unique.
     """
 
+    @classmethod
+    def _recursive_add(cls, g: List["Model"]) -> List["Model"]:
```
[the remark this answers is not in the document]
erichaydel: Done!

ormar/relations/relation_proxy.py
erichaydel: Done!

collerek (approved):

PR #971, opened 2022-12-27
fix(tests): TestClient no longer accepts json argument for get() and delete() methods

sandrotosi:


sandrotosi:
@collerek can you elaborate why you closed this PR without merging nor commenting?

collerek:
Sorry, it's no longer valid as we switched to AsynClient from httpx in tests instead of TestClient from starlette, you can check master branch (and 0.20.0 pre-release with support for pydantic v2)

PR #985, opened 2023-01-12
feat: sort choices

zevisert:
We want our generated OpenAPI file to generate the same way each time (docker build cache concerns). We use a couple hacks to get that to happen, including:

- `random.seed(...)` before importing any ormar models to prevent variability from the foreign key's `PkOnlyModelXXXXX` and get_pydantic's `Model_XXX` generated classes
- `freezegun.freeze_time()` before instantiating any models that use a default factory for a datetime
- hard-coding UUIDs with `UUID('<uuid>')` instead of calling `uuidv4()`

Yet still there is variability with enum choice order that we cannot control externally. So, this patch adds a model field kwarg to sort choices and store them as a list instead.


ormar/fields/model_fields.py
```diff
@@ -68,7 +68,8 @@ def convert_choices_if_needed(
     nullable: bool,
     scale: int = None,
     represent_as_str: bool = False,
-) -> Set:
+    sort_choices: bool = True,
+) -> List:
```
zevisert, line 72: This should typecheck since `BaseField` types choices as a Sequence instead of a concrete type like list or set

PR #989, opened 2023-01-19
allow index into ForeignKey

cmflynn:


cmflynn:
> Thanks, seems right to include it! :)
> 
> Can you please add test (check underlying sqlalchemy table structure if it has the index set / not set based on this setting - you can check Model.Meta.table for columns settings and table indexes) and docs for it (in fk section)

@collerek I looked through the docs and wasn't sure exactly where/what we should add this for argument. The foreign key section does not actually document any of the `kwargs` that are curried from ormars fk to sqlalchemys fks. I suppose we could start there? happy to do that, or if you have a more specific suggestion of where this documentation should go I'm happy to add it.

collerek:
> 

That would be awesome!
Yep the fk section looks like a good candidate for this.

collerek:
Merged in https://github.com/collerek/ormar/pull/1276 as I don't have perms on your fork, thanks!

Issue #1136, opened 2023-07-10
Add `first_or_none`


dekoza:
**Is your feature request related to a problem? Please describe.**
I'm coming from Django and I'm used to having `first()` return None if there are no matches. Ormar should either replicate this behavior or introduce `first_or_none` method to go in sync with other `_or_none` methods. 

Side note: Documentation for `get()` and `get_or_none()` is misleading as it reads "Returns **first** row" which sadly is not true because it raises `MultipleMatches` if there are more rows matching criteria.

**Describe the solution you'd like**
I already have a PR adding `first_or_none` but the docs suggest opening an issue beforehand :wink: 

Side note: mypy's pre-commit hook yields a lot of errors from other code in the repo.

PR #1137, opened 2023-07-10
Add QuerySet.first_or_none

dekoza:
As I mentioned in #1136 - here's the code providing missing `first_or_none` method.

I also update docstrings to better match what each method actually raises.


ormar/queryset/queryset.py
```diff
@@ -894,7 +894,6 @@ async def first(self, *args: Any, **kwargs: Any) -> "T":
         Gets the first row from the db ordered by primary key column ascending.
 
         :raises NoMatch: if no rows are returned
-        :raises MultipleMatches: if more than 1 row is returned.
```
collerek, line 897: Why did you remove it - it does indeed check for one row only in `self.check_single_result_rows_count(processed_rows)`, please restore it.

ormar/queryset/queryset.py
```diff
@@ -924,6 +923,22 @@ async def first(self, *args: Any, **kwargs: Any) -> "T":
         self.check_single_result_rows_count(processed_rows)
         return processed_rows[0]  # type: ignore
 
+    async def first_or_none(self, *args: Any, **kwargs: Any) -> Optional["T"]:
+        """
+        Gets the first row from the db ordered by primary key column ascending.
+
+        If no match is found None will be returned.
+
+        :param kwargs: fields names and proper value types
```
collerek, line 932: This as a result can also raise MultipleMatches, please add:
` :raises MultipleMatches: if more than 1 row is returned.`

collerek (changes requested):

collerek:
Sorry for the long time of review :/

collerek (approved):

Issue #789, opened 2022-08-19
Joins result in an error with tables that have a period in their name


pmdevita:
**Describe the bug**

https://github.com/collerek/ormar/blob/a27d7673a5354429bef4158297d76a58522c1579/ormar/queryset/join.py#L112

Because `SqlJoin._on_clause` expects a `from_clause` string in the format of "table_name.column_name", if your table name has a period in it, the string will be chunked into more than two parts, causing an error when it cannot set the two variables.

**To Reproduce**
1. Create a model with a foreign key and a tablename with a period in it
2. Attempt to filter data over the foreign key relation, which should result in a join and then in an error

```ValueError: too many values to unpack (expected 2) ```

**Expected behavior**
Ormar handles the period without issue.

**Screenshots**

**Versions (please complete the following information):**
 - Database backend used: MySQL and SQLite
 - Python version: 3.10
 - `ormar` version: 0.11.2
 - `pydantic` version: 1.9.1

**Solution**

One solution I would propose since users generally don't use periods in their columns is to just reserve the last chunk for the column name and do something like this.

```
parts = from_clause.split(".")
table = ".".join(parts[:-1])
column = parts[-1]
```

However, a more robust solution would be to just keep table name and column name separate strings until they are run through `quotter()`.

Let me know if there is a preferred method to solve this and I'll open a pull request.

pmdevita:
This should be completed with the merge of #1174, closing

PR #1174, opened 2023-08-24
fix: Table names and columns not being quoted properly (#789)

pmdevita:
When Ormar was performing a left join, it would not correctly escape the table and column name, occasionally causing trouble with some characters. This fixes quoting when it generates a left join with an alias, and when ordering with that alias.

### DIFF
--- a/ormar/queryset/join.py
+++ b/ormar/queryset/join.py
@@ -92,7 +92,7 @@ class SqlJoin:
         """
         return self.next_model.Meta.table
 
-    def _on_clause(self, previous_alias: str, from_table_name:str, from_column_name: str, to_clause: str) -> text:
+    def _on_clause(self, previous_alias: str, from_table_name:str, from_column_name: str, to_table_name: str, to_column_name: str) -> text:
         """
         Receives aliases and names of both ends of the join and combines them
         into one text clause used in joins.
@@ -108,11 +108,11 @@ class SqlJoin:
         :return: clause combining all strings
         :rtype: sqlalchemy.text
         """
-        left_part = f"{self.next_alias}_{to_clause}"
+        dialect = self.main_model.Meta.database._backend._dialect
+        quoter = dialect.identifier_preparer.quote
+        left_part = f"{quoter(self.next_alias + '_' + to_table_name)}.{quoter(to_column_name)}"
         if not previous_alias:
-            dialect = self.main_model.Meta.database._backend._dialect
-            quotter = dialect.identifier_preparer.quote
-            right_part = f"{quotter(from_table_name)}.{quotter(from_column_name)}"
+            right_part = f"{quoter(from_table_name)}.{quoter(from_column_name)}"
         else:
             right_part = f"{previous_alias}_{from_table_name}.{from_column_name}"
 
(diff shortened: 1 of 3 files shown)


pmdevita:
This fixes #789 as well.

ormar/queryset/join.py
```diff
@@ -92,28 +92,29 @@ def to_table(self) -> sqlalchemy.Table:
...
+        :type from_column_name: str
         :param to_clause: to table name
         :type to_clause: str
         :return: clause combining all strings
         :rtype: sqlalchemy.text
         """
-        left_part = f"{self.next_alias}_{to_clause}"
+        dialect = self.main_model.Meta.database._backend._dialect
+        quoter = dialect.identifier_preparer.quote
+        left_part = f"{quoter(self.next_alias + '_' + to_table_name)}.{quoter(to_column_name)}"
```
collerek, line 113: Please replace string concat with nested f-string
`left_part = f"{quoter(f'{self.next_alias}_{to_table_name}')}.{quoter(to_column_name)}"`

ormar/queryset/join.py
```diff
@@ -92,28 +92,29 @@ def to_table(self) -> sqlalchemy.Table:
...
         if not previous_alias:
-            dialect = self.main_model.Meta.database._backend._dialect
-            table, column = from_clause.split(".")
-            quotter = dialect.identifier_preparer.quote
-            right_part = f"{quotter(table)}.{quotter(column)}"
+            right_part = f"{quoter(from_table_name)}.{quoter(from_column_name)}"
         else:
-            right_part = f"{previous_alias}_{from_clause}"
+            right_part = f"{previous_alias}_{from_table_name}.{from_column_name}"
```
collerek, line 117: Shouldn't this also be quoted? 🤔

tests/test_model_definition/test_field_quoting.py
```diff
@@ -0,0 +1,75 @@
...
+    class2 = await SchoolClass.objects.create(name="Logic")
+    category = await Category.objects.create(name="Foreign")
+    category2 = await Category.objects.create(name="Domestic")
+    await Student.objects.create(name="Jane", category=category, schoolclass=class1)
+    await Student.objects.create(name="Judy", category=category2, schoolclass=class1)
+    await Student.objects.create(name="Jack", category=category2, schoolclass=class2)
+
+
+@pytest.mark.asyncio
+async def test_quotes_left_join():
```
collerek, line 67: Please also add tests for reverse relation joins (so i.e. coming from schoolclass -> students) and deeper joins (like schoolclass -> students -> categories)

collerek (commented):
Thanks!

One minor change and some missing tests and we are good to go :)

ormar/queryset/join.py
pmdevita: Probably so, added!

ormar/queryset/join.py
pmdevita: Ah I thought that was going to be a 3.12 thing lol

pmdevita:
Done!

collerek (approved):
Awesome, thanks!

PR #1197, opened 2023-09-18
Update dependencies

TouwaStar:
Hey, i would love if we could update databases to newest version as we are a couple bugfixes behind, let me know how can we get this merged.


collerek (approved):
Awesome, thx!

PR #1224, opened 2023-11-04
Update README.md

wesleylima:
ormar is no longer supported by fastapi-users :(
https://github.com/fastapi-users/fastapi-users/blob/master/docs/migration/9x_to_10x.md#tortoise-orm-and-ormar


collerek (approved):
Thanks!

PR #1249, opened 2024-01-19
Getting test suites to pass

TouwaStar:
Just going through pytest with every test folder for now

Test suites **working now:**
test_inheritance_and_pydantic_generation
test_meta_constraints
test_utils
test_relations
test_queries
test_ordering
test_model_methods
test_model_definitions
test_exclude_include_dict
test_encryption
test_deferred

**TODO**:
test_fastapi


TouwaStar:
@collerek can you review and see if we can get it into the pydantic 2 branch?

ormar/models/newbasemodel.py
```diff
@@ -144,15 +144,16 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:  # type: ignore
         new_kwargs, through_tmp_dict = self._process_kwargs(kwargs)
 
         if not pk_only:
-            self.__pydantic_validator__.validate_python(
+            values = self.__pydantic_validator__.validate_python(
                 new_kwargs, self_instance=self  # type: ignore
-            )
+            ).__dict__
+            fields_set = set(values.keys())
```
collerek, line 150: I am almost sure it's initializing self (so values and fields_set) in ` self.__pydantic_validator__.validate_python` as it worked in tests w/o your change (that's why you pass self), now you do this 2 times (with this setattr below)

ormar/models/newbasemodel.py
```diff
@@ -208,11 +206,6 @@ def __getattr__(self, item: str) -> Any:
             return None
         return super().__getattr__(item)
 
-    def __getattribute__(self, item: str) -> Any:
-        if item == "__dict__":
```
collerek, line 212: yeah that was to check if we can hijack __dict__ calls from typeadapter in fastapi, can be removed but this is one of the ideas how to work around this.

tests/test_fastapi/test_binary_fields.py
```diff
@@ -3,6 +3,7 @@
 import uuid
 from enum import Enum
 from typing import List
+import codecs
```
collerek, line 6: unused import?

tests/test_fastapi/test_binary_fields.py
```diff
@@ -37,7 +38,7 @@ async def shutdown() -> None:
         await database_.disconnect()
 
 
-blob3 = b"\xc3\x28"
+blob3 = b"\xc3\x83\x28"
```
collerek, line 41: why this change?

TouwaStar: The utf was invalid, there is no c3 or c3 28.
It was failing on trying to decode it because of that.
C3 83 is greek capital A 
https://www.cogsci.ed.ac.uk/~richard/utf-8.cgi?input=00C2&mode=hex

TouwaStar:
I think the circular reference can be avoided if we add special handling for reverse type relationships

ormar/relations/relation.py
```diff
@@ -155,6 +155,9 @@ def add(self, child: "Model") -> None:
         if self._type in (RelationType.PRIMARY, RelationType.THROUGH):
             self.related_models = child
             self._owner.__dict__[relation_name] = child
+        elif self._type == RelationType.REVERSE:
+            # TODO: Define what to do with reverse relations
+            pass
```
collerek, line 160: The reverse relation logic is include in else now. The change in pydantic v2 is that instead of calling a method to dump to json/dict it's now calling rust code which gets the whole object and basically grabs the `__dict__` directly, that's the check in `__getattribute__` but that happens also in other cases. Before ormar used to keep related models on relation/relation proxy and dump them to dict only on dict/json calls. Now I had to add the related models to __dict__ too to ensure proper dumping in pydantic. That seems to work well except for the cases in which we have circular references. As there is no exact point to remove them before dumping (before there was a relationship map in ormar that was built on dump call that excluded circular references). 

The whole problem is how to maintain access to nested models and keep the relations accessible in code, but exclude them in dumping. So, when you have models `A->B` you can still get the `B` and dump it in a way that it has the `A` in related models without the additional db call, but at the same time prevent dumping `A->B->A`.

ormar/fields/parsers.py
```diff
@@ -33,15 +33,19 @@ def encode_decimal(value: decimal.Decimal, precision: int = None) -> float:
 
 def encode_bytes(value: Union[str, bytes], represent_as_string: bool = False) -> bytes:
     if represent_as_string:
-        return value if isinstance(value, bytes) else base64.b64decode(value)
-    return value if isinstance(value, bytes) else value.encode("utf-8")
+        value=  value if isinstance(value, bytes) else base64.b64decode(value)
+    else:
+        value = value if isinstance(value, bytes) else value.encode("utf-8")
+    print(' encode_bytes: value =', value)
```
collerek, line 39: Are those prints still needed? (lines 39 & 48)

ormar/relations/relation.py
TouwaStar: Pydantic offers a solution for serializing recursive fields but only on a field by field case if i understand correctly 

https://github.com/pydantic/pydantic/blob/b2aa36a80904773d7db2e4c007fc00a5a4d7653b/docs/concepts/postponed_annotations.md?plain=1#L251

If we could make this kind of "Skip if pydantic thinks its repeating" logic for all model fields by default we could escape this but it doesnt seem like a nice solution

collerek: Nice catch! didn't see it before :) 

IIUC it would have to be only the reverse relation field and not all of them. It might be kind of ugly but hey, there are already some ugly workarounds in ormar anyway 🤣 I want to use the v2 to do some refactoring if possible, but the primary goal is to make it work with v2.

PR #1252, opened 2024-01-25
Fixup some tests

TouwaStar:


tests/test_fastapi/test_wekref_exclusion.py
```diff
@@ -157,6 +157,5 @@ async def test_endpoints():
         assert resp5.status_code == 200
         assert len(resp5.json()) == 3
 
-        resp6 = await client.get("/test/error")
-        assert resp6.status_code == 200
-        assert len(resp6.json()) == 3
+        with pytest.raises(ReferenceError):
```
TouwaStar, line 160: @collerek  can you verify im thinking right here ? I imagine this should be raising thats why i changed it

collerek: I just fixed that, it's excluded in response_model so should be fine, it was some check done in rust that should have not been done if that was excluded, just fixed all tests on local 🎉

TouwaStar: Nice!!

PR #1259, opened 2024-02-01
Lifespan

mekanix:
It's hard with tests as when first fails, some of the next ones fail, too. I checked them individually, and every test I changed passes like that. That probably means I need better `lifespan()` but for now I just wanted to show my intention and ask if I'm going in the right way?

### DIFF
--- a/.github/workflows/test-package.yml
+++ b/.github/workflows/test-package.yml
@@ -39,7 +39,7 @@ jobs:
           POSTGRES_DB: testsuite
         ports:
           - 5432:5432
-        options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
+        options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5 --name postgres
 
     steps:
     - name: Checkout
@@ -70,6 +70,30 @@ jobs:
       env:
         DATABASE_URL: "mysql://username:password@127.0.0.1:3306/testsuite"
       run: bash scripts/test.sh
+
+    - name: Install postgresql-client
+      run: |
+        sudo apt-get update
+        sudo apt-get install --yes postgresql-client
+
+    - name: Connect to PostgreSQL with CLI
+      run: env PGPASSWORD=password psql -h localhost -U username -c 'SELECT VERSION();' testsuite
+
+    - name: Show max connections
+      run: env PGPASSWORD=password psql -h localhost -U username -c 'SHOW max_connections;' testsuite
+
+    - name: Alter max connections
+      run: |
+        
+        docker exec -i postgres bash << EOF
+          sed -i -e 's/max_connections = 100/max_connections = 1000/' /var/lib/postgresql/data/postgresql.conf
+          sed -i -e 's/shared_buffers = 128MB/shared_buffers = 512MB/' /var/lib/postgresql/data/postgresql.conf
+        EOF
+        docker restart --time 0 postgres
+        sleep 5
+
+    - name: Show max connections
+      run: env PGPASSWORD=password psql -h localhost -U username -c 'SHOW max_connections;' testsuite
 
     - name: Run postgres
       env:
--- a/benchmarks/conftest.py
+++ b/benchmarks/conftest.py
@@ -3,26 +3,18 @@ import random
 import string
 import time
 
-import databases
 import nest_asyncio
 import ormar
 import pytest
 import pytest_asyncio
-import sqlalchemy
-from tests.settings import DATABASE_URL
 
+from tests.settings import create_config
+from tests.lifespan import init_tests
+
+
+base_ormar_config = create_config()
 nest_asyncio.apply()
-
-
-database = databases.Database(DATABASE_URL)
-metadata = sqlalchemy.MetaData()
 pytestmark = pytest.mark.asyncio
-
-
-base_ormar_config = ormar.OrmarConfig(
-    metadata=metadata,
-    database=database,
-)
 
 
 class Author(ormar.Model):
@@ -57,13 +49,7 @@ class Book(ormar.Model):
     year: int = ormar.Integer(nullable=True)
 
 
-@pytest.fixture(autouse=True, scope="function")  # TODO: fix this to be module
-def create_test_database():
-    engine = sqlalchemy.create_engine(DATABASE_URL)
-    metadata.drop_all(engine)
-    metadata.create_all(engine)
-    yield
-    metadata.drop_all(engine)
+create_test_database = init_tests(base_ormar_config, scope="function")
 
 
 @pytest_asyncio.fixture
--- a/docs/fastapi/index.md
+++ b/docs/fastapi/index.md
@@ -26,7 +26,8 @@ Here you can find a very simple sample application code.
 
 ### Imports and initialization 
 
-First take care of the imports and initialization 
+Define startup and shutdown procedures using FastAPI lifespan and use is in the
+application.
 ```python
 from typing import List, Optional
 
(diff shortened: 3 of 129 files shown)


collerek:
Yep, that seems like the way to go. Note also that `app.state` is deprecated so it should be like global variables (https://github.com/tiangolo/fastapi/issues/81#issuecomment-473677039) with db instead of custom properties, or something like a dependency with cache (https://fastapi.tiangolo.com/advanced/settings/#creating-the-settings-only-once-with-lru_cache).

collerek:
Ah almost forgot, I would like to switch to async tests for fastapi too, that means going away from TestClient from starlette and switching to httpx. So we need httpx in dev dependencies, and a client fixture in each file, something like this:

```python
@pytest_asyncio.fixture()
async def client(request: pytest.FixtureRequest) -> AsyncIterator[AsyncClient]:
    transport = ASGITransport(app=app, client=("1.2.3.4", 123), raise_app_exceptions=False)
    async with AsyncClient(transport=transport) as client, LifespanManager(app):
        yield client
```

Note that for tests and lifetime, you need an additional dev dependency `asgi_lifespan` with `LifespanManager` like used above and AsyncClient from httpx. Then switch all requests to async using this client like `await client.get(...)` (and tests need to be marked with `@pytest.mark.asyncio`).

gmekicaxcient:
One thing at a time. First I need lifespan to work. If I put database instance as global var, all tests register their models and of course, there are name collisions. I need to drop database or drop all tables in tear down, first. Let me experiment with it for a bit and once it works, async will be easier.

collerek:
Sure thing! :)

Btw by global, I meant scope -> so global to the module/file not one for all tests (global for the app). 
So each lifespan refers to its own global `_database = Database(...)` somewhere below the imports in each of the files.

mekanix:
That would be nice, but then I can't reach `_database` from `lifespan`. As I said, let me experiment a little and see how to do this.

collerek:
Not sure what you mean you cannot access it -> sample of what I aim for is here: https://github.com/collerek/ormar/blob/a8b9fcea0cde3cdc6b619961134c410d6fea8406/tests/test_fastapi/test_binary_fields.py just changed to lifespan. Realized httpx and asgi)lifespan deps are already there too, forgot about this 🤣

mekanix:
I didn't fix all the tests, but I think you can see what I want to do. I intend to finish what I've started tomorrow, unless you have some objections/comments.

collerek:
Yeah it seems fine, will save a lot of duplication. Not sure if we need engine in the OrmarConfig as it's used only in tests to create/ drop db (in production one should use migrations) and is not used anywhere in the ormar itself. But at the same time it's optional so could probably just sit there if anyone wants it, need to think about it 🤔

mekanix:
I squashed all commits and made all tests use new functions `create_config`, `lifespan` and `init_tests`. Benchmarks are failing, but they were failing even before I changed anything. As a matter of fact, some of the benchmarks pass now, which is a progress. Anyway, I can't figure out why the two tests fail
```
FAILED tests/test_fastapi/test_binary_fields.py::test_read_main - AssertionError: assert 'w4Mo' == 'Ã('
FAILED tests/test_fastapi/test_json_field_fastapi.py::test_setting_values_after_init - sqlite3.OperationalError: table things has no column named js
```

Can you please look into that? Second failure is strange, because if you run only that test, it passes, but I can't figure out where I'm not cleaning what I should clean up.

As for the `engine` being part of config, I would say it's a good thing to have it there. As ormar is sqlaclhemy consumer, and every sqlaclhemy config consists of `database`, `metadata` and `engine`, I think this makes it complete and exposes `engine` for developers if they ever need it. For tests it's obvious why we need it, as we need to drag it along with `database` and `metadata` somehow. Also, if tests are any indication of possible use case, other people will need `engine`, too.

collerek:
You need `create_test_database = init_tests(base_ormar_config)` in each of the files as there are models with duplicated table names. When you add it to `test_json_field_fastapi.py` file it passes, binary test also has model with table named things that's why those two conflict. Please add this line to all of the files so they are self contained (it can cause hard to track bugs in other cases, each file is considered a stand alone tests.)

mekanix:
Ah, great catch! It was already 4am when I sent this PR, no wonder I missed it. I'll fix in few hours when I finish some chores. Thank you!

mekanix:
I initialized all tests, but I still get two errors
```
FAILED tests/test_fastapi/test_binary_fields.py::test_read_main - AssertionError: assert 'w4Mo' == 'Ã('
ERROR tests/test_model_methods/test_populate_default_values.py::test_populate_default_values - sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) near "Default": syntax error
```
Any clues? I can't see what's wrong, honestly.

collerek:
I modified the workflows to run also tests again pydantic_v2 branch, can you push something here, I guess it should be picked up. Cause I run the tests on local after checking out your pr and all did pass for me. If the gh will not pick up the changes because v2 branch is not merged yet then you should be able to change it here I guess but let's see.

mekanix:
Ah, I was speaking too soon. I only see MySQL failing, so I'll try to setup MySQL test on my desktop.

collerek:
mysql is first, then pg and sqlite at the end. You can use docker to test it -> you have sample docker-compose and commands in makefile

collerek:
but it seems you forgot lifetime method in test_fastapi with nested defaults, as for the model methods tests I believe it should be `text("'Default Name'")` with double quotes, not sure how that passes now 😅

mekanix:
We have a saying in Serbian (which is almost the same as Croatian) "you and I make a good one" :o) With the latest error, I guess it's some PostgreSQL throttling?

collerek:
that's a bit weird - you did add some create calls but was it that many?
We can test the increased number of connections, add 
`command: postgres -c 'max_connections=1000'` in `.github/workflows/test-package.yml`
so it will become:
```yaml
      postgres:
        image: postgres:10.8
        command: postgres -c 'max_connections=1000'
        env:
          POSTGRES_USER: username
          POSTGRES_PASSWORD: password
          POSTGRES_DB: testsuite
        ports:
          - 5432:5432
        options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
```

collerek:
Ah shit command is not supported in github actions :/ You need to follow the steps here: https://stackoverflow.com/questions/70673766/how-to-increase-max-connection-in-github-action-postgres

collerek:
it's `--host` instead of `--hostname` https://www.postgresql.org/docs/current/app-psql.html

collerek:
You are missing db name, check https://stackoverflow.com/questions/32824388/postgresql-remotely-connecting-to-postgres-instance-using-psql-command

mekanix:
I don't know if I would ever catch that. Nice! Now looking at next error, how is postgresql supposed to be restarted via docker if we're inside docker already?

collerek:
Add name to options so container have static name instead of random docker one and use that one. You can follow the guide here: https://github.com/remarkablemark/remarkablemark.github.io/blob/master/_posts/2022/2022-05-12-github-actions-postgresql-increase-max-connections-and-shared-buffers.md

mekanix:
Are we OK with this regression in benchmark tests?

collerek (approved):
Great job! That was a lot more changes than I expected (didn't think you would refactor to remove duplication for all tests) but more than welcome! Saved me a ton of work and I really appreciate it! Thank you! :) 

As for the benchmarks - yeah degradation is fine, I still need to figure out if they are reliable enough :)

mekanix:
First of all, thank you for the great ORM! Second of all, thank you for the kind words! Third of all, what should I do next? The two that stand out are ` Code cleanup` and `Fix line numbers highlighted in docs examples after reformat`. When I say stand out, I mean the rest of the requirements sound like I need way more familiarity with ormar code than I currently have. Please do correct me if I'm wrong and advise on the next step(s).

collerek:
If you would like to help more the `Fix line numbers highlighted in docs examples after reformat` is a good candidate. 😁 

When you open docs of the current version there are highlighted some lines in the code examples. i.e line 10 below
image

(screen comes from https://collerek.github.io/ormar/models/)

If you checkout wip_pydantic_v2 branch and issue `mkdocs serve` you will see that now this line is line 9 not 10.

image

So in file `docs/models/index.md` the lines:

image

Should be corrected to `Python hl_lines="9"`

That needs to be checked/corrected for all code samples in the docs (and it's  a nice way to get to know ormar better)

mekanix:
```
grep -r hl_lines | wc -l
      59
```
Not bad! I'll take it.

PR #1265, opened 2024-02-14
Change line numbers in documentation

mekanix:
I changed all occurrences of `Meta` that I knew how to, as well. Line numbers change when I replace `Meta` with `ormar_config`, so I did that. If you grep, you'll find more `Meta` instances around, but I don't know how to change them. For example, I never worked with signals, so I don't know how's that implemented now.

PR #1267, opened 2024-02-19
Fix documentation

wor1dedit:


collerek:
Hi, thanks for the help! But it overlaps with part of the https://github.com/collerek/ormar/pull/1265 that's why I asked to write in the pr thread if you want to take something on 😅 Can you sync with @mekanix on who is doing what? What would you like to update?

mekanix:
If I'm correct, the mentioned PR already has all these changes. As in the following weeks I'll be busy elsewhere, maybe @wor1dedit can work on `Update docs to reflect breaking changes` from the list? It's just a suggestion.

PR #1273, opened 2024-02-27
feat: add on_update to the field

vvanglro:
This is a refinement of the #532.

### DIFF
--- a/ormar/models/model.py
+++ b/ormar/models/model.py
@@ -119,6 +119,7 @@ class Model(ModelRow):
         ):
             await self.load()
 
+        self.__setattr_fields__.clear()
         await self.signals.post_save.send(sender=self.__class__, instance=self)
         return self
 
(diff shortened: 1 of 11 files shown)


ormar/models/model.py
```diff
@@ -242,7 +243,19 @@ async def update(self: T, _columns: List[str] = None, **kwargs: Any) -> T:
...
         self_fields.pop(self.get_column_name_from_alias(self.Meta.pkname))
         if _columns:
-            self_fields = {k: v for k, v in self_fields.items() if k in _columns}
+            onupdate_fields = self.get_fields_with_onupdate()
+            self_fields = {
+                k: v
+                for k, v in self_fields.items()
+                if k in _columns or k in onupdate_fields
+            }
+        if not kwargs and not _columns:
```
vvanglro, line 252: If all are null here, the update is based on user-modified attributes on the instance, with user-modified attributes taking precedence over on_update.

PR #1302, opened 2024-05-14
Fix add_field_serializer_for_reverse_relations clearing validators

camillol:
When a model has a reverse relation, Ormar synthesizes a field for it. In order to get Pydantic to see the new field, it calls `DecoratorInfos.build` (in `add_field_serializer_for_reverse_relations`) to pick up the new serializer, and then `model_rebuild`. Unfortunately, `DecoratorInfos.build` is not idempotent. When it processes a model class, it collects the information in `__pydantic_decorators__`, and it strips out the decorators from the class members, to save time in case it has to process the class again (as the base class of a different class). If it is called again, it starts collecting the decorators from scratch, ignoring the existing `__pydantic_decorators__`. Since the class members have already had their decorators stripped, they are not collected again. The net result is that any pydantic decorators that had already been processed are lost.

This has the result of ignoring `@field_validator` etc. if a model has a reverse relation that is processed after the model.

This PR fixes the issue by storing the preexisting `__pydantic_decorators__` and merging the new ones into it. This fix should keep working even if Pydantic changes `DecoratorInfos.build` to be idempotent.


collerek:
sorry for the late answer, can you add a test that would fail before and this change fixes?

camillol:
> sorry for the late answer, can you add a test that would fail before and this change fixes?

Done!

collerek:
Hi, you didn't allow me to push to your fork, so can you please apply https://github.com/collerek/ormar/commit/b1d803a945687d37e6101e981615daaacbf859f5 on your pr (fixes test with default value)

camillol:
The missing coverage is due to a block that was not changed in this PR. Seems to be a preexisting issue.

collerek:
> The missing coverage is due to a block that was not changed in this PR. Seems to be a preexisting issue.

Yeah I know but since I cannot edit your pr can you add a `  # pragma: no cover` to not cover this block so we can merge it? Also you need to modify `@pytest_asyncio.fixture(scope="function")` and add `autouse=True` to this as the cleanup fixture is not used now since you didn't refer to it in test nor added the autouse param.

camillol:
Done, and you should be allowed to edit it now.

collerek (approved):
Awesome, thanks!

Issue #1381, opened 2024-07-13
Random error when prefetching models


cadlagtrader:
**Describe the bug**
I am seeing random errors when prefetching a model with ormar 0.20.1
Identified the bug in    ```translate_list_to_dict``` using a dict as a default argument

**To Reproduce**
```translate_list_to_dict(["aa", "aa__inner", "bb"], default={})```

**Expected behavior**
ret = {"aa": {"inner": {}}, "bb": {}}

cadlagtrader:
[https://github.com/collerek/ormar/pull/1382](url)

PR #1382, opened 2024-07-13
copy translate_list_to_dict default argument

cadlagtrader:
Fixes [https://github.com/collerek/ormar/issues/1381](https://github.com/collerek/ormar/issues/1381)

### DIFF
--- a/ormar/queryset/utils.py
+++ b/ormar/queryset/utils.py
@@ -64,7 +64,7 @@ def translate_list_to_dict(  # noqa: CCR001
     for path in list_to_trans:
         current_level = new_dict
         parts = path.split("__")
-        def_val: Any = default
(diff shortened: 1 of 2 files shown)


ormar/queryset/utils.py
```diff
@@ -64,7 +64,7 @@ def translate_list_to_dict(  # noqa: CCR001
     for path in list_to_trans:
         current_level = new_dict
         parts = path.split("__")
-        def_val: Any = default
+        def_val: Any = copy.copy(default)
```
collerek, line 67: I would go with deepcopy as it can be nested dict, list of dicts or some mutable custom object

cadlagtrader: Makes sense. Updated and added test

collerek (approved):

PR #1388, opened 2024-08-01
Allow Pydantic >= 2.5.3

camillol:


collerek:
That's still installing `2.5.3` in tests, might be cross dependency with pydantic-extra or something else. As it mingles with pydantic internals each upper version should be checked.

camillol:
Ran poetry lock!

camillol:
@collerek there is one broken test, `tests/test_fastapi/test_binary_fields.py::test_read_main`. It's not clear to me that the behavior expected by the test is correct; it looks like it wants the JSON response to represent the LargeBinary field as its binary contents decoded into a string with the default encoding (utf-8), but why should it? Could you take care of it?

PR #1395, opened 2024-09-18
Support python 3.12

ChristopherMacGown:
- **typing: remove types-pkg-resources package**
- **fix: recursive_guard is a KW_ONLY arg in 3.12.4+**


pyproject.toml
```diff
@@ -55,6 +55,7 @@ asyncpg = { version = ">=0.28,<0.30", optional = true }
 psycopg2-binary = { version = "^2.9.1", optional = true }
 mysqlclient = { version = "^2.1.0", optional = true }
 PyMySQL = { version = "^1.1.0", optional = true }
+setuptools = "^75.1.0"
```
collerek, line 58: Why do we need this?

ChristopherMacGown: CodeFactor failed the prior commit due to https://github.com/advisories/GHSA-cx63-2mw6-8hw5

ChristopherMacGown:
Regarding the failed test, `distutils` was deprecated in Python 3.10 and removed in Python 3.12. The test job pins poetry to 1.4.2, but the first version of poetry that supports 3.12 is 1.7.0. I would bump the poetry version in the PR, but I noticed that the version is pinned to different versions in different workflows and that don't want to bump them all without a discussion.

collerek:
> Regarding the failed test, `distutils` was deprecated in Python 3.10 and removed in Python 3.12. The test job pins poetry to 1.4.2, but the first version of poetry that supports 3.12 is 1.7.0. I would bump the poetry version in the PR, but I noticed that the version is pinned to different versions in different workflows and that don't want to bump them all without a discussion.

I guess we can bump it and unify the version, note that you will have to relock the file with new version probably.

ChristopherMacGown:
I added Python 3.13 to the test matrix to try and get coverage there but it looks like there aren't ryu wheels for it yet. I'll push up a commit removing it soon

ChristopherMacGown:
The [CodeFactor](https://www.codefactor.io/repository/github/collerek/ormar/pull/1395) report is about a vulnerability in multipart/form-data in starlette versions prior to 0.40.0. In order to clear that report and update starlette, the fastapi pin will need to be relaxed to allow 0.115.3 to be installed, currently it is pinned to versions below 0.112.0.

collerek (approved):
Awesome, thanks a lot! :)

PR #1420, opened 2024-12-10
Allow running with pydantic 2.9

hauntsaninja:


collerek:
Added 2.11 in https://github.com/collerek/ormar/pull/1470

PR #1445, opened 2025-03-20
chore: 🔧 modify All Fields return type

Sube-py:
For IDE type hints, e.g. str type, you can hint both splits and functions on String.


collerek (approved):
Awesome, thanks!

PR #1493, opened 2026-01-26
Added schema field to OrmarConfig

hk3dva:
Added use of schema in foreign keys, Many to Many
Added field to schema conversion via sqlalchemy

### DIFF
--- a/.github/workflows/test-package.yml
+++ b/.github/workflows/test-package.yml
@@ -61,6 +61,13 @@ jobs:
     - name: Install dependencies
       run: poetry install --extras "all"
 
+    - name: Grant MySQL global privileges
+      # The schema-per-model tests create extra databases for cross-schema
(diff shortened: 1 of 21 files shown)


collerek:
Hi, looks promising but I'm missing some more query-related tests -> there is quite a complicated table and column prefixing happening under the hood to prevent colliding on the same names between tables and columns. 

That's why we need to test some joins -> like select related, prefetch related, same table joins, multiple joins to the same table, accessing relationships and reverse relationships from models, and some filtering. 

Between the tables in different schemas as same table schema is covered, probably also some joins to the 2 tables with same name but in different schemas. I know that it's most common use case is separation and not combining data in different schemas, but with proper permissions, that is a perfectly valid use case, so we need to make sure it works as expected.

collerek (approved):

PR #1580, opened 2026-03-10
feat: add choices to strings to enable literal types

### LINKED ISSUE #1579: Updating type annotations to support string literals
**Is your feature request related to a problem? Please describe.**
Using the `ormar.String` with a string literal type hint causes a type error, as the ormar string is more permissive than the literal:

```py
mode: Literal["user", "manager", "admin"] = ormar.String(max_length=32)
# results in: Type "str" is not assignable to declared type "Literal['user', 'manager', 'admin']"
```

**Describe the solution you'd like**
Ideally, either a new model field, `ormar.Literal`, or the ability to provide valid values to the `ormar.String` field, thus removing the type error.

**Describe alternatives you've considered**
I have tried using enums before, however they can be tedious to work with in Postgres due to the addition of new values requiring a hefty migration (which can be very slow to run against large datasets). Using a varchar results in it being more flexible, but it lacks this degree of type safety.

Another option is to use an ormar string with a string type hint, then cast the value to the literal once it's confirmed to be a valid value, however this pushes a lot of boilerplate into the domain code. Having the model field typed as a literal immediately means it can be consumed from anywhere and be guaranteed to be correct.

**Additional context**
Nil.

MaximSrour:
## Summary

Enables `ormar.String` to preserve string literal typing when `choices` are provided, while keeping the underlying database column as `sqlalchemy.String`.

Closes #1579.

## Changes

- update `ormar.String` to project a `typing.Literal[...]` pydantic type when string `choices` are provided
- preserve nullable behavior for `String(..., choices=..., nullable=True)` by wrapping the projected type in `Optional[...]`
- keep normal `String` behavior unchanged when `choices` are not provided
- add/adjust tests for:
  - runtime validation of `String(..., choices=...)`
  - nullable string choices
  - generated JSON schema for string choices fields
- update field documentation to describe `String` and `String(..., choices=...)` in a single section

## Testing

Built and ran targetted tests for the new changes, and ran tests against the entire suite. All passing.

Installed a local version of ORMAR into a personal repository that uses it, and tested against existing type errors.

## Screenshots

Error before:
image

Using the new literal inference:
image

### DIFF
--- a/docs/fields/field-types.md
+++ b/docs/fields/field-types.md
@@ -19,10 +19,32 @@ Each of the `Fields` has assigned both `sqlalchemy` column class and python type
         regex: str = None,)` has a required `max_length` parameter.  
 
 * Sqlalchemy column: `sqlalchemy.String`  
-* Type (used for pydantic): `str` 
+* Type (used for pydantic): `str`, or `typing.Literal[...]` when string `choices` are provided
 
 !!!tip
     For explanation of other parameters check [pydantic](https://pydantic-docs.helpmanual.io/usage/schema/#field-customisation) documentation.
+
+You can also constrain string values with `choices` while keeping a varchar column:
+
+```python
+from typing import Literal
+
+import ormar
+
+
+class Account(ormar.Model):
+    ormar_config = ...
+
+    id: int = ormar.Integer(primary_key=True)
+    mode: Literal["user", "manager", "admin"] = ormar.String(
+        max_length=32,
+        choices=("user", "manager", "admin"),
+    )
+```
+
+For the strongest static typing, keep the explicit field annotation. If you rely on
+inference from the `choices` argument alone, type checkers are more likely to infer a
+`str` type instead of the appropriate literal.
 
 ### Text
 
@@ -219,7 +241,6 @@ So which one to use depends on the backend you use and on the column/ data type 
 
 * Sqlalchemy column: `sqlalchemy.Enum`  
 * Type (used for pydantic): `type[Enum]`
-
 
 [relations]: ../relations/index.md
 [queries]: ../queries.md
--- a/ormar/fields/model_fields.py
+++ b/ormar/fields/model_fields.py
@@ -3,7 +3,7 @@ import decimal
 import uuid
 from enum import Enum as E
 from enum import EnumMeta
-from typing import Any, Optional
+from typing import Any, Literal, Optional
 
 import pydantic
 import sqlalchemy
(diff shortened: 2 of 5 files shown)


collerek:
Any particural reason why you dont want to use EnumField here?

MaximSrour:
> Any particural reason why you dont want to use EnumField here?

I wrote about it a bit in the linked issue, but it mostly comes down to flexibility - with an enum, I'd need to push a migration to update the column type (which can take a while when there are a lot of records), after which I'll be able to use the new value. With a string column instead, the code can immediately start using the new values.

collerek:
We used to have `choices` for all field types that was removed in https://github.com/ormar-orm/ormar/releases/tag/0.20.0 

If we would like to restore it as `Literal` type I guess that should support all field types again and not only string one. After all `Literal[1,2,3]` is also a valid literal and could be used to limit Integer field.

Note also that you now overwrite the `overwrite_pydantic_type` that the user might have provided, along with choices - which at least require a warning displayed.

MaximSrour:
> We used to have `choices` for all field types that was removed in https://github.com/ormar-orm/ormar/releases/tag/0.20.0

Ah, it was removed a bit before I had started using ORMAR, so I was never aware that it was even a feature in the first place. I imagine that it was removed for a good reason, so it may be unwise to just add it back in. That's fine though, this PR was just a quick attempt to see if it was an quick and easy fix.

> should support all field types again and not only string one

I noted that `typing` exports a `LiteralString`, which might be a valid alternative, though I haven't much experience with it. I might fiddle around with it to see if it works.

Unless you mean that the `choices` field is added back in for **all** model fields, in which case then there wouldn't be a point to the above investigation (beyond curiousity).

> you now overwrite the `overwrite_pydantic_type` that the user might have provided

Ah, interesting, I'd need to investigate this, but good catch.

Either way, I'm fine with this PR (and the linked issue) being closed, since the desired feature was removed in the first place. Ultimately up to you.

collerek:
To be honest I dont remember why it was removed now 😅 

Might have been an issue with the implementation not working with pydantic 2.0. It was not using the literals for sure. 

I guess we could restore it - as the lack of migration is a valid benefit for that use case, but as I mentioned in the comment, then we should make it generic for all column types in ormar so choices can be applied also on ints or other types.

MaximSrour:
Lol, no worries. Gimme a couple of weeks to find the time, and I'll get this one resolved 🤞

PR #1637, opened 2026-04-22
refactor(queryset): use sqlalchemy.column() for SelectAction aggregate identifiers

tomaioo:
## Summary

Security: Potential SQL injection via raw SQL text construction in select clause

## Problem

**Severity**: `High` | **File**: `ormar/queryset/actions/select_action.py:L38`

The `SelectAction.get_text_clause()` method builds a SQL fragment with `sqlalchemy.text(f"{alias}{self.field_name}")`. If `select_str` can be influenced by untrusted input (for example, API-driven dynamic field selection), this pattern can allow SQL injection through crafted field names or aliases because raw text is concatenated directly into SQL.

## Solution

Avoid `sqlalchemy.text()` for identifiers derived from strings. Resolve fields strictly through model metadata (e.g., `table.c[column_name]`) after whitelist validation, and reject unknown/non-identifier field names. If textual SQL is unavoidable, enforce a strict regex such as `^[A-Za-z_][A-Za-z0-9_]*$` on each identifier component before use.

## Changes

- `ormar/queryset/actions/select_action.py` (modified)

### DIFF
--- a/ormar/queryset/actions/query_action.py
+++ b/ormar/queryset/actions/query_action.py
@@ -48,7 +48,7 @@ class QueryAction(abc.ABC):
     @abc.abstractmethod
     def get_text_clause(
         self,
-    ) -> sqlalchemy.sql.expression.TextClause:  # pragma: no cover
+    ) -> sqlalchemy.sql.expression.ClauseElement:  # pragma: no cover
         pass
 
     @property
--- a/ormar/queryset/actions/select_action.py
+++ b/ormar/queryset/actions/select_action.py
@@ -1,6 +1,5 @@
 import decimal
-import re
-from typing import TYPE_CHECKING, Any, Callable, Optional
+from typing import TYPE_CHECKING, Any, Callable
 
 import sqlalchemy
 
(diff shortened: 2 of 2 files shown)


collerek:
Thanks for the report @tomaioo — I don't think this is reachable though, so I pushed a follow-up commit narrowing the PR.                                                                                                             
                                                                                                                                                                                                                                        
  `SelectAction.get_text_clause()` is only called from `_query_aggr_function` (`queryset/queryset.py`), which already whitelists `field_name` against `target_model.model_fields` and raises `QueryDefinitionError` before the `text()` 
  call — covered by `tests/test_vulnerabilities/test_aggregated_functions.py`. `table_prefix` is never user-controlled: it's either empty or an internal alias from `AliasManager.get_table_alias()` (two uppercase + four hex chars).
  So the regex guard and `ValueError` raises duplicate validation that already exists one frame up.                                                                                                                                     
                                                                    
  The `sqlalchemy.text()` → `sqlalchemy.column()` swap is still a nice cleanup on its own, so I kept that and dropped the rest:                                                                                                         
   
  - Removed the regex/ValueError blocks and `import re` (redundant).                                                                                                                                                                    
  - Removed the unused `alias` kwarg on `__init__` and the `Optional` import.
  - Widened `QueryAction.get_text_clause` return type to `ClauseElement` for LSP; corrected `apply_func` return type.                                                                                                                   
                                                                                                                                                                                                                                        
  Retitling to reflect that this is a refactor, not a security fix. Full test suite still passes.

collerek:
BTW the security vulnerabilities should be reported in Security Advisory to get maintainers a chance to patch it before making it public

PR #1741, opened 2026-07-15
docs: add complete alembic integration example

### LINKED ISSUE #1400: A *working* alembic example in the docs
The [docs regarding alembic integration](https://collerek.github.io/ormar/latest/models/migrations/#alembic-usage) are too laconic to be considered useful for beginners. They look more like reminders for folks that have at least 5 years of alembic experience under their belt.

I wish the examples provided there would have all imports in place. E.g. where I should take `metadata` from? How does that `my_models` file looks like if I have the metadata exposed there?

If I have several apps with models in them - what then? What should I prepare to feed the alembic config?

There are [some tutorials in the wild](https://pawamoy.github.io/posts/testing-fastapi-ormar-alembic-apps/) but they use old ormar (pre 0.20). They can be ported with a little work but I think such examples should be a part of the official docs.

pctablet505:
Fixes #1400.

The current Alembic docs skip too many steps for anyone new to ormar — imports, where `metadata` comes from, and how to handle models split across apps. I added a complete, working example based on the FastAPI-template reference linked in the issue.

It covers a single-file layout, a multi-app layout, a minimal `alembic.ini`, and wiring the shared `metadata` into `alembic/env.py`. I ran `alembic revision --autogenerate` and `alembic upgrade head` against both layouts and built the docs with `mkdocs build` to make sure everything works.

### DIFF
--- a/docs/models/migrations.md
+++ b/docs/models/migrations.md
@@ -48,55 +48,13 @@ alembic upgrade head
 
 Ormar models are built on SQLAlchemy Core. Each model stores its table on a SQLAlchemy `MetaData` instance that you provide through `OrmarConfig`. Alembic needs that same `MetaData` object so it can discover which tables exist and autogenerate migrations.
 
-The usual pattern is to create one shared `MetaData` object and pass it to every model:
-
-```python
-# my_project/models.py
-import sqlalchemy
-import ormar
-from ormar import DatabaseConnection
-
-database = DatabaseConnection("sqlite+aiosqlite:///db.sqlite")
-metadata = sqlalchemy.MetaData()
-
-
-class Author(ormar.Model):
-    ormar_config = ormar.OrmarConfig(
-        database=database,
-        metadata=metadata,
-        tablename="authors",
-    )
-
-    id: int = ormar.Integer(primary_key=True)
-    name: str = ormar.String(max_length=100)
-
-
-class Book(ormar.Model):
-    ormar_config = ormar.OrmarConfig(
-        database=database,
-        metadata=metadata,
-        tablename="books",
-    )
-
-    id: int = ormar.Integer(primary_key=True)
-    title: str = ormar.String(max_length=100)
-    author: Author = ormar.ForeignKey(Author)
-```
-
-You can then expose `metadata` from the module that defines your models:
-
-```python
-# env.py
-from my_project.models import metadata
-
-target_metadata = metadata
-```
-
-If you prefer, you can also grab the metadata from any model:
+The usual pattern is to create one shared `MetaData` object and pass it to every model in its `OrmarConfig`. You can then expose that metadata from any model:
 
 ```python
 target_metadata = Author.ormar_config.metadata
 ```
+
+or export the shared `metadata` instance from your models module and import it into `env.py`.
 
 The important part is that every model uses the **same** `sqlalchemy.MetaData()` instance. If models live in different files or apps, import them all before Alembic inspects the metadata (see [Multiple apps with models](#multiple-apps-with-models) below).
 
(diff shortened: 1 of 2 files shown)


docs/models/migrations.md
```diff
@@ -38,102 +38,181 @@ Likewise as with tables, since we base tables on sqlalchemy for migrations pleas
...
+    id: int = ormar.Integer(primary_key=True)
+    title: str = ormar.String(max_length=100)
+    author: Author = ormar.ForeignKey(Author)
+```
 
-    This configures the context with just a URL
-    and not an Engine, though an Engine is acceptable
-    here as well.  By skipping the Engine creation
-    we don't even need a DBAPI to be available.
+`alembic.ini` (only the parts you typically need to change):
```
collerek: Copying this file together with the `env.py` below gives:

```
File ".../logging/config.py", line 115, in _create_formatters
    flist = cp["formatters"]["keys"]
KeyError: 'formatters'
```

`env.py` calls `fileConfig(config.config_file_name)`, which needs `[loggers]`, `[handlers]` and `[formatters]` — none of which are here.

The mixed framing is what causes it. The heading says "Complete project layout", the intro says "minimal but complete layout that works with `alembic revision --autogenerate`", and `env.py` is given in full as a drop-in replacement — so it's natural to read this file the same way. But the caption says "only the parts you typically need to change".

Either include the generated logging sections so the file is genuinely complete, or reword to something unambiguous: "In the generated `alembic.ini`, change only these keys — leave the rest, including the logging sections, as-is."

For what it's worth, with the logging sections restored both layouts work end to end: autogenerate detects `authors` and `books`, and `alembic upgrade head` creates them. So the substance is right, it's just the ini that's under-specified.

docs/models/migrations.md
```diff
@@ -38,102 +38,181 @@ Likewise as with tables, since we base tables on sqlalchemy for migrations pleas
...
-    Calls to context.execute() here emit the given string to the
-    script output.
+```ini
+[alembic]
+script_location = %(here)s/alembic
 
-    """
+prepend_sys_path = .
+
+sqlalchemy.url = sqlite:///db.sqlite
```
collerek: This is correct, but the doc never says why it differs from the `sqlite+aiosqlite:///db.sqlite` the models use a few blocks up. Given the audience the issue describes, someone will "fix" the inconsistency and land on:

```
sqlalchemy.exc.MissingGreenlet: greenlet_spawn has not been called;
can't call await_only() here.
```

which is not a message you can search your way out of. Could you add a short callout right here — ormar talks to the database through an async driver, alembic's default `env.py` is synchronous, so the two URLs point at the same database through different drivers on purpose. A small table would carry a lot of weight:

| database | ormar (`DatabaseConnection`) | alembic (`sqlalchemy.url`) |
|---|---|---|
| SQLite | `sqlite+aiosqlite:///db.sqlite` | `sqlite:///db.sqlite` |
| PostgreSQL | `postgresql+asyncpg://...` | `postgresql+psycopg2://...` |
| MySQL | `mysql+aiomysql://...` | `mysql+pymysql://...` |

And one line pointing at `alembic init -t async` for anyone who'd rather keep a single URL.

docs/models/migrations.md
```diff
@@ -38,102 +38,181 @@ Likewise as with tables, since we base tables on sqlalchemy for migrations pleas
...
+│   └── models.py
+├── alembic.ini
+└── db.sqlite
+```
 
-# add your model's MetaData object here (the one used in ormar)
-# for 'autogenerate' support
-from app.models.my_models import metadata
-target_metadata = metadata
+`my_project/models.py`:
```
collerek, line 78: We keep runnable doc examples in `docs_src/` and pull them into the markdown with the snippet syntax — see `docs/signals.md:38` for the pattern. `docs_src/test_all_docs.py` executes every `.py` under that tree as a script in CI, so those examples can't silently rot.

Could these two model modules move to e.g. `docs_src/models/docs0NN.py` and be included with:

```
--8<-- "../docs_src/models/docs0NN.py"
```

`scripts/test_docs.sh` only runs `pytest docs_src/` — markdown code fences are never executed, which is exactly how the `alembic.ini` problem above got through a green CI run. `env.py` and `alembic.ini` obviously can't live there; inline is fine for those.

docs/models/migrations.md
```diff
@@ -38,102 +38,181 @@ Likewise as with tables, since we base tables on sqlalchemy for migrations pleas
...
 
-```python
+```bash
 alembic init alembic
 alembic revision --autogenerate -m "made some changes"
 alembic upgrade head
 ```
 
-### Sample env.py file
+### Where does `metadata` come from?
```
collerek, line 47: The `Author`/`Book` definitions here are byte-identical to the ones in "Complete project layout" ~70 lines down. The `# env.py` snippet in this section also imports `from my_project.models import metadata` before `my_project` has been introduced anywhere.

Suggest trimming this section to the prose plus the one-liner it already has:

```python
target_metadata = Author.ormar_config.metadata
```

and letting "Complete project layout" carry the single full listing. The explanation stands on its own without 30 lines of models repeated.

docs/models/migrations.md
```diff
@@ -38,102 +38,181 @@ Likewise as with tables, since we base tables on sqlalchemy for migrations pleas
...
-        url=URL,
+        url=url,
         target_metadata=target_metadata,
         literal_binds=True,
         dialect_opts={"paramstyle": "named"},
-        # if you use UUID field set also this param
-        # the prefix has to match sqlalchemy import name in alembic
-        # that can be set by sqlalchemy_module_prefix option (default 'sa.')
-        user_module_prefix='sa.'
+        # Required if you use ormar.UUID().
```
collerek: The old comment carried the part that's actually actionable: the prefix has to match the SQLAlchemy import name used in the generated migration, which alembic exposes as `sqlalchemy_module_prefix` (default `sa.`). Anyone who changed that setting now has no way to know what belongs here. Could you keep a condensed version of the *why*?

Unrelated but adjacent: the `compare_type` section further down still has `'sa.'` in single quotes while everything new uses double — worth normalizing since you're already touching quoting in this file.

docs/models/migrations.md
```diff
@@ -38,102 +38,181 @@ Likewise as with tables, since we base tables on sqlalchemy for migrations pleas
...
+`alembic.ini` (only the parts you typically need to change):
 
-    Calls to context.execute() here emit the given string to the
-    script output.
+```ini
+[alembic]
+script_location = %(here)s/alembic
 
-    """
+prepend_sys_path = .
```
collerek: This and the `sys.path.insert(0, str(Path(__file__).resolve().parents[1]))` in `env.py` do the same job. Having both is confusing for exactly the reader this doc is aimed at. I'd keep the `env.py` one — it works regardless of which directory alembic is invoked from — and drop this line.

Related: both `sqlalchemy.url` and the ormar URL are cwd-relative, so `db.sqlite` only lands at the project root shown in the tree if you always run from the root. `sqlite:///%(here)s/db.sqlite` pins it.

docs/models/migrations.md
```diff
@@ -144,9 +223,106 @@ if context.is_offline_mode():
...
+```
+
+Then generate and run your first migration:
+
+```bash
+alembic revision --autogenerate -m "initial"
+alembic upgrade head
+```
+
+### Multiple apps with models
```
collerek, line 211: `authors/` and `books/` have no `__init__.py` here, while `my_project/` and `models/` do. It works via namespace packages (I checked), but the inconsistency reads as an oversight in a doc whose entire purpose is leaving nothing implicit.

Also: the single-file `env.py` puts `from my_project.models import metadata` after `sys.path.insert`, which is E402 under most linters. The multi-app snippet carries `# noqa: F401` but the single-file one has nothing — consider `# noqa: E402` on both so people don't paste a lint error into their project.

collerek (changes requested):

pctablet505:
Hi @collerek, thanks for the review! All 7 points have been addressed in the branch:

1. **Snippet inclusion**: Extracted models to `docs_src/models/docs019.py` and included them via `--8<-- "../docs_src/models/docs019.py"`.
2. **Trimmed duplicates**: Simplified "Where does `metadata` come from?" to the single-line extraction.
3. **Logging sections**: Added standard `[loggers]`, `[handlers]`, and `[formatters]` to `alembic.ini` so `fileConfig` works out of the box.
4. **Driver comparison**: Added a callout + driver comparison table explaining async vs. sync URLs (`aiosqlite`/`asyncpg` vs. `sqlite`/`psycopg2`) and pointed to `alembic init -t async`.
5. **Path handling**: Dropped `prepend_sys_path` from `alembic.ini` (handled by `env.py`) and pinned `sqlite:///%(here)s/db.sqlite`.
6. **Package layout & lint**: Added missing `__init__.py` files to the multi-app layout and `# noqa: E402` to `env.py`.
7. **Module prefix**: Retained context explaining `user_module_prefix="sa."` must match `sqlalchemy_module_prefix`, and normalized quote formatting.

Branch is rebased onto the latest `master` and mergeable. Ready for your re-review whenever you have time.

